PR #238, opened 2015-02-27
APPS-1235 - Not Crashing on Messages With a Single Message Part Image.

chipxsd:
Fixes the crash where trying to initialize `ATLMessageCollectionViewCell` with a frame that has a `NaN` value for height.

Android sample app doesn't yet send multipart messages (with thumbnails and image size info) along with the full resolution photos.

### DIFF
--- a/Tests/ATLMediaAttachmentTests.m
+++ b/Tests/ATLMediaAttachmentTests.m
@@ -41,13 +41,6 @@ NSData *ATLTestAttachmentDataFromStream(NSInputStream *inputStream);
  */
 UIImage *ATLTestAttachmentMakeImageWithSize(CGSize size);
 
-/**
- @abstract Synhchronously grabs the last photo from the Photos Library.
- @param library The library to grab the last photo from.
- @return Returns ALAsset instance of the last image located in the Photos Library, or `nil` in case of a failure.
- */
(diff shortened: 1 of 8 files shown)


Tests/ATLMediaAttachmentTests.m
```diff
@@ -0,0 +1,307 @@
...
+    
+    // Verifying image metadata JSON
+    NSData *imageSizeMetadataJSON = ATLTestAttachmentDataFromStream(mediaAttachment.metadataInputStream);
+    expect(imageSizeMetadataJSON).toNot.beNil();
+    expect(imageSizeMetadataJSON).to.equal([NSJSONSerialization dataWithJSONObject:@{ @"width":@1024, @"height":@512, @"orientation":@(UIImageOrientationUp) } options:NSJSONWritingPrettyPrinted error:nil]);
+}
+
+@end
+
+#pragma mark - Test utilities
```
blakewatters: Can we please move these to the top of the file or create a `ATLTestUtilities.{h,m}` file? I think it’s valuable to be able to jump to the bottom of the file and just start appending tests

chipxsd: Can do.

PR #733, opened 2015-03-01
use full width in input view if left accessory button is set to nil

grevolution:
Added check for messageInputView TextView to use full width if leftAccessoryButton is set to nil


Code/Views/ATLMessageInputToolbar.m
```diff
@@ -96,10 +96,13 @@ - (void)layoutSubviews
     CGRect leftButtonFrame = self.leftAccessoryButton.frame;
     CGRect rightButtonFrame = self.rightAccessoryButton.frame;
     CGRect textViewFrame = self.textInputView.frame;
-
-    leftButtonFrame.size.width = ATLLeftAccessoryButtonWidth;
-    leftButtonFrame.size.height = ATLButtonHeight;
-    leftButtonFrame.origin.x = ATLLeftButtonHorizontalMargin;
+    
+    //only to calculate when leftAccessroyButton is available
+    if(self.leftAccessoryButton){
```
blakewatters: Spacing: 
```objc
// Only to calculate when leftAccessroyButton is available
if (self.leftAcessoryButton) {`
```

Code/Views/ATLMessageInputToolbar.m
```diff
@@ -116,9 +119,12 @@ - (void)layoutSubviews
     frame.size.height = CGRectGetHeight(textViewFrame) + ATLVerticalMargin * 2;
     frame.origin.y -= frame.size.height - CGRectGetHeight(self.frame);
  
+    //check if leftAccessroyButton is available.
```
blakewatters: Spacing on the comment as well

grevolution:
the CI build failed because of rake version according to logs. Do I have to do something about it?

kcoleman731:
@grevolution Looks like this branch needs a rebase now. Care to address the spacing comments from @blakewatters  and push it up?

kcoleman731:
@grevolution We added this functionality to Atlas. It is on master and will be in our next release.

PR #739, opened 2015-03-05
Update ATLAvatarItem.h

danipralea:
Add protocol for Asynchronous image load


kcoleman731:
Hi @danipralea We just added this functionality to Atlas. It is on master.

Issue #746, opened 2015-03-09
NSInvalidArgumentException when leaving a conversation


jamiefaye:
Using Atlas Messenger, create a conversation with several participants. Then go into the detail view and click on "Leave Conversation". This causes the failure to occur most of the time.

The offending call is in -    - ATLUIConversationViewController.m configureControllerForChangedParticipants.

Here is the backtrace:
2015-03-09 11:57:54.476 Atlas Messenger[635:181238] -[__NSOrderedSetI minusSet:]: unrecognized selector sent to instance 0x17001d880
2015-03-09 11:57:54.476 Atlas Messenger[635:181238] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSOrderedSetI minusSet:]: unrecognized selector sent to instance 0x17001d880'
*** First throw call stack:
(0x18579e59c 0x195ee80e4 0x1857a5664 0x1857a2418 0x1856a6b6c 0x1000fa410 0x1000f9bec 0x185744ae4 0x185683220 0x186582c60 0x1001dc02c 0x100710e30 0x100710df0 0x10071575c 0x185755fa4 0x18575404c 0x1856810a4 0x18e81b5a4 0x189fb2aa4 0x1000cefa4 0x196556a08)
libc++abi.dylib: terminating with uncaught exception of type NSException
(lldb) bt
* thread #1: tid = 0x2c3f6, 0x000000019666f270 libsystem_kernel.dylib`__pthread_kill + 8, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x000000019666f270 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000019670d228 libsystem_pthread.dylib`pthread_kill + 112
    frame #2: 0x00000001965e6b18 libsystem_c.dylib`abort + 112
    frame #3: 0x00000001956b9418 libc++abi.dylib`abort_message + 116
    frame #4: 0x00000001956d8b8c libc++abi.dylib`default_terminate_handler() + 304
    frame #5: 0x0000000195ee83c0 libobjc.A.dylib`_objc_terminate() + 128
    frame #6: 0x00000001956d5bb4 libc++abi.dylib`std::__terminate(void (*)()) + 16
    frame #7: 0x00000001956d5478 libc++abi.dylib`__cxa_throw + 136
    frame #8: 0x0000000195ee8204 libobjc.A.dylib`objc_exception_throw + 348
[...]

PR #748, opened 2015-03-09
fixed method that took in NSOrderedSet instead of NSSet

kmahal:
fixing this issue: https://github.com/layerhq/Atlas-iOS/issues/746

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -748,7 +748,7 @@ static NSString *const ATLPushNotificationSoundName = @"layerbell.caf";
         [self configureConversationForAddressBar];
         return;
     }
-    NSMutableSet *removedParticipantIdentifiers = [self.typingParticipantIDs copy];
+    NSMutableSet *removedParticipantIdentifiers = [NSMutableSet setWithObject:[self.typingParticipantIDs mutableCopy]];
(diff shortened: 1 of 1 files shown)


Code/Controllers/ATLConversationViewController.m
```diff
@@ -748,7 +748,7 @@ - (void)configureControllerForChangedParticipants
         [self configureConversationForAddressBar];
         return;
     }
-    NSMutableSet *removedParticipantIdentifiers = [self.typingParticipantIDs copy];
+    NSMutableSet *removedParticipantIdentifiers = [NSMutableSet setWithObject:[self.typingParticipantIDs mutableCopy]];
```
kcoleman731: `self.typingParticipantIDs` is already an NSMutableOrderedSet. Calling the following returns another mutable ordered set so is not necessary. 
```
[self.typingParticipantIDs mutableCopy]
```
You are also assigning `removedParticipantIdentifiers` to be an NSMutableSet which contains one object, an `NSMutableOrderedSet`. Don't think that is what you want correct?

You probably want the following...
```
[NSMutableSet setWithArray:[self.typingParticipantIDs array]];
```
The array call on an NSMutableOrderedSet returns a `facade` object which you can then use to create the Mutable set. 

Oh collections!

PR #751, opened 2015-03-13
Update README.md

sanvean:
When adding Atlas to project directly rather than using CocoaPods, 2 additional frameworks are required than are currently listed in the README


kcoleman731:
Thank you @sanvean

PR #757, opened 2015-03-16
Allow hiding edit button & search bar in conversations list

melvynhills:
- `displaysEditButton` can be toggled before & after the view is loaded.
- `displaysSearchBar ` can only be configured before the view is loaded (mimicking `displaysAddressBar` from `ATLBaseConversationViewController`).

### DIFF
--- a/Code/Controllers/ATLConversationListViewController.h
+++ b/Code/Controllers/ATLConversationListViewController.h
@@ -169,6 +169,16 @@
 @property (nonatomic, assign) BOOL allowsEditing;
 
 /**
+ @abstract IA boolean value to determine whether or not the receiver should display an Edit button made available with `allowsEditing`. Default is yes.
+  */
+@property (nonatomic, assign) BOOL displaysEditButton;
+
+/**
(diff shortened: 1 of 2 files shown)


kcoleman731:
@melvynhills Thank you for the PR. I think these are very reasonable additions. Want to add a few tests for em and we can get it merged?

kcoleman731:
Hi @melvynhills. Never heard back from you on this PR?

I'm going to go ahead and close this as it is actually unnecessary. The `searchBar` property of the `ATLConversationListViewController` is set as `self.tableView.tableHeaderView`. If you subclass `ATLConversationListViewController`, you can just set the tableHeaderView property to `nil` if you don't want a search bar displayed. 

You can also set the `leftAccessoryButton` to `nil` if you don't want to see the edit button. 

Fair enough?

melvynhills:
Fair enough. Thanks for the alternative tricks.  I guess it's something more specific to my needs anyway. Didn't have the time to figure out the whole testing system either.

PR #758, opened 2015-03-16
Fix default value for displaysAddressBar, as per documentation

melvynhills:
As per documentation, displaysAddressBar should default to `NO`.
The superfluous `_displaysAddressBar = YES` is now removed.


kcoleman731:
@melvynhills Nice catch. Merged.

PR #793, opened 2015-04-21
Update ATLAvatarImageView.m

danipralea:
Without this, it used to crash in my project if the ```imageURL``` object was not nil, but empty


kcoleman731:
Thanks @danipralea

PR #794, opened 2015-04-21
displaysAvatarItem out of sync with documentation

andrewcopp:
Changed documentation of displaysAvatarItem to NO because initialization defaulted to NO. I assume that the documentation was supposed to change and not the code because this way will more clearly mirror iOS Messages app.


andrewcopp:
Yikes. Didn't check to see if this would break your tests.

kcoleman731:
Thank you @andrewcopp!

PR #795, opened 2015-04-21
cornerRadius calculations for ATLAvatarImageView

andrewcopp:
ATLAvatarImageView wasn't working for custom components. The corner radius was set to a literal value. By keying off of the frame of the image view instead, we were able to get a more dynamic calucalation. Still room for improvement but this works and doesn't break anything going on in ATLConversationTableViewCell for example

### DIFF
--- a/Code/Views/ATLAvatarImageView.m
+++ b/Code/Views/ATLAvatarImageView.m
@@ -70,7 +70,7 @@ NSString *const ATLAvatarImageViewAccessibilityLabel = @"ATLAvatarImageViewAcces
     // Default UI Appearance
     _initialsFont = [UIFont systemFontOfSize:14];
     _initialsColor = [UIColor blackColor];
-    _avatarImageViewDiameter = 27;
+    _avatarImageViewDiameter = CGRectGetHeight(self.frame);
     
     self.clipsToBounds = YES;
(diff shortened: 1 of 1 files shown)


kcoleman731:
@andrewcopp  The diameter property is actually meant to be configure via `UIAppearanceProperty`. It is documented the the default diameter is 30 (I need to correct the actual implementation).  
```
@property (nonatomic) CGFloat avatarImageViewDiameter UI_APPEARANCE_SELECTOR;
```
Your suggestion to use `CGRectGetHeight(self.frame);` will have no effect if the view is not initialized with a frame.

andrewcopp:
Good catch. What about configuring the `cornerRadius` in `layoutSubviews` where you will have access to the frame? If the documentation says it's a circle view, having to configure the diameter seems like an extra step.

kcoleman731:
Hey @andrewcopp my apologies for the delay on this. Trying to catch up on some Atlas stuff today. 

Does using the public property not suit your use case? If so, can you tell me (explain the problem a bit more) why and we can figure out the ideal solution?

Thanks!

andrewcopp:
I think got it working using the public property. I can't remember my thought process after rereading my last comment so let's close it.

PR #798, opened 2015-04-23
Check fetched image prior to adding to cache

chadpod:
When I have a poor network connection I am seeing a lot of exceptions here, where image is nil.


kcoleman731:
Thanks Chad!

PR #807, opened 2015-04-27
Sent date instead of received date in conversation list

piotrbernad:


piotrbernad:
Second change is about clearing text from input toolbar. There are some situations when it shouldnt be cleared for example when there is no coversation to send message.

kcoleman731:
@piotrbernad Thank you for this PR!

We sort the conversation list according to the conversations `lastMessage.receivedAt` property. If you would like the flexibility to change the sorting and also the date displayed in the cells, I would be very open to a PR that provides that functionality. 

Also, your PR has changes to the ruby version file and the podfile.lock which we wouldn't want to merge.

kmahal:
Hey @piotrbernad any update on this PR?  If not we can close it out.  It also must be re-based since it's not mergeable anymore.

kmahal:
@piotrbernad I'll be closing this out.  Feel free to re-open if you update this.

PR #808, opened 2015-04-27
Thumbnail Size in input view

piotrbernad:
Added posibility to modify thumbnail size of image attachment when presented in input toolbar.


chipxsd:
Looks solid, thanks.

chipxsd:
Looks solid, merged!

PR #810, opened 2015-04-28
conversation list storyboard support

andrewcopp:
The ATLConversationListViewController doesn't currently support the use of Storyboards. It forces the table cell reuse identifier on any subclasses which break the registration set up by the storyboard. This pull request exposes this API as well as removes some other superfluous code.


kcoleman731:
@andrewcopp  This a a really great addition and something I've wanted to add for a bit. Thank you for the PR.

PR #812, opened 2015-04-29
photo action sheet

andrewcopp:
property for the photo action sheet prevents photo selection from appearing when developers add additional action sheets

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -57,6 +57,7 @@ static NSString *const ATLDefaultPushAlertGIF = @"sent you a GIF.";
 static NSString *const ATLDefaultPushAlertImage = @"sent you a photo.";
 static NSString *const ATLDefaultPushAlertLocation = @"sent you a location.";
 static NSString *const ATLDefaultPushAlertText = @"sent you a message.";
+static NSInteger const ATLPhotoActionSheet = 1000;
 
 + (instancetype)conversationViewControllerWithLayerClient:(LYRClient *)layerClient;
 {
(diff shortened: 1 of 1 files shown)


Code/Controllers/ATLConversationViewController.m
```diff
@@ -494,6 +496,7 @@ - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTap
                                                     cancelButtonTitle:@"Cancel"
                                                destructiveButtonTitle:nil
                                                     otherButtonTitles:@"Take Photo", @"Last Photo Taken", @"Photo Library", nil];
+    [self setPhotoActionSheet:actionSheet];
```
kmahal: You can set the `UIActionsheet` initializer here directly to `self.photoActionSheet` instead of setting the pointer to the created instance since the property is explicitly only used here.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -602,21 +605,23 @@ - (void)sendMessageWithLocation:(CLLocation *)location
...
-           [self captureLastPhotoTaken];
-            break;
-          
-        case 2:
-            [self displayImagePickerWithSourceType:UIImagePickerControllerSourceTypePhotoLibrary];
-            break;
-            
-        default:
-            break;
+    if ([actionSheet isEqual:[self photoActionSheet]]) {
```
kmahal: use `self.photoActionSheet` here to follow the coding style

kcoleman731:
@andrewcopp rather than creating a new property, this same effect can be achieved via the `tag` property on the Action sheet. Would you be opposed to using that?

andrewcopp:
I suppose that's the intended use of the tag property. It would differentiate the action sheets without having to store one in a property. :+1:

andrewcopp:
This fell off of my radar. Will revisit next week when I return.

kmahal:
Hey @andrewcopp any update on this one?  Cleaning up the repo and trying to close out open items.

andrewcopp:
@kmahal we should be good to go!

kmahal:
Hey @andrewcopp slight detail, but we like to avoid merging master into our branches, since we rebase instead.  Rebase your last commit on top of master and we can get this guy merged down.  Thanks!

andrewcopp:
@kmahal should be good now.

kmahal:
@andrewcopp looks good.  merged.

PR #813, opened 2015-05-01
Feature/apps 1548 atl conversation list view controller data source method

kmahal:
This PR creates a data source method to let users define the last message text for custom MIMEtypes.

### DIFF
--- a/Code/Controllers/ATLConversationListViewController.h
+++ b/Code/Controllers/ATLConversationListViewController.h
@@ -210,7 +210,7 @@
 /**
  @abstract The controller used to display search results.
  */
-@property (nonatomic, readonly) UISearchDisplayController *searchContorller;
+@property (nonatomic, readonly) UISearchDisplayController *searchController;
 
 
 ///------------------------------
--- a/Code/Controllers/ATLConversationListViewController.h
+++ b/Code/Controllers/ATLConversationListViewController.h
@@ -109,9 +109,9 @@
 - (UIColor *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController colorForButtonWithDeletionMode:(LYRDeletionMode)deletionMode;
 
 /**
- @abstract Asks the data source for the text string to display as the conversation's last sent message.
- @params conversation The conversation to return the last message text.
- @return A string for the last sent message.  If `nil` is returned the controller will fall back to default behavior.
+ @abstract Asks the data source for the string to display as the conversation's last sent message.
+ @params conversation The conversation for which the last message text should be returned.
+ @return A string representing the content of the last message.  If `nil` is returned the controller will fall back to default behavior.
  @discussion This is used when the application uses custom `MIMEType`s and wants to customize how they are displayed.
  */
 - (NSString *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController lastMessageTextForConversation:(LYRConversation *)conversation;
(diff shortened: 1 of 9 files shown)


Code/Controllers/ATLConversationListViewController.h
```diff
@@ -108,6 +108,14 @@
  */
 - (UIColor *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController colorForButtonWithDeletionMode:(LYRDeletionMode)deletionMode;
 
+/**
+ @abstract Asks the data source for the text string to display as the conversation's last sent message.
```
kcoleman731: `text string` is redundant.

Code/Controllers/ATLConversationListViewController.h
```diff
@@ -108,6 +108,14 @@
...
+ @params conversation The conversation to return the last message text.
+ @return A string for the last sent message.  If `nil` is returned the controller will fall back to default behavior.
```
kcoleman731: The conversation for which the last message text should be returned.

kcoleman731: A string representing the content of the last message.

Code/Views/ATLConversationTableViewCell.h
```diff
@@ -21,6 +21,22 @@
 #import <UIKit/UIKit.h>
 #import "ATLConversationPresenting.h"
 
+@class ATLConversationTableViewCell;
+
+@protocol ATLConversationTableViewCellDataSource <NSObject>
```
kcoleman731: We use the presenting pattern all over Atlas so that controllers can `inform` cells of their content. It also allows custom cells to conform to the presenter protocols if they wish to implement their own subclass. We should add a method to the `ATLConversationPresenting` protocol rather than add a datasource.

kmahal: @kcoleman731 I was thinking about that, but that's already taken care in custom cells in `ATLConversationPresenting`:
```objc
- (void)presentConversation:(LYRConversation *)conversation;
```
For custom cells, the user can handle the conversation, and it's `lastMessage` as they see fit.  This data source method is used for non-custom cells, but where they send custom `MIMEType`s in the `ATLConversationViewController` (and subsequently have custom `UICollectionViewCell` subclasses they register in that controller).

kcoleman731: Understood. What we should probably do then is break out the default logic that determine the last message text and move it into the controller. TBH, that is probably where it belongs anyways. So then you keep the `presentConversation` method, because people def will want that, but we add the `updateWithLastMessageText:`. This mirrors what we have for conversation title.

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -265,6 +269,16 @@ - (void)configureCell:(UITableViewCell<ATLConversationPresenting> *)conversation
     } else {
         @throw [NSException exceptionWithName:NSInternalInconsistencyException reason:@"Conversation View Delegate must return a conversation label" userInfo:nil];
     }
+    
+    NSString *lastMessageText;
+    if ([self.dataSource respondsToSelector:@selector(conversationListViewController:lastMessageTextForConversation:)]) {
+        lastMessageText = [self.dataSource conversationListViewController:self lastMessageTextForConversation:conversation];
+    }
+    if (lastMessageText == nil) {
```
kcoleman731:
```
if (!lastMessageText) {
     lastMessageText = [self defaultLastMessageTextForConversation:conversation]
}
[conversationCell updateWithLastMessageText:lastMessageText];
```
I think this is more readable...less logic.

kcoleman731:
@kmahal lets add a test to explicitly exercise the new `ATLConversationPresenting` method and this bad boy will be good to merge.

kmahal:
@kcoleman731 existing `ATLConversationTableViewCellTest` class already exercises `ATLConversationPresenting` to ensure the lastMessage label is correct, which now directly draws from the newly added method, so no new test is required.

PR #821, opened 2015-05-08
Fix incorrect if statement

mikewu1211:
The if statement always returns “false” because [self.message valueForKeyPath:@"parts.MIMEType"] is an array.


kmahal:
@mikewu1211 this was an older dependency when Android only sent 1 message part for images with the specific `MIMEType`.  We compare that array to another array `@[ATLMIMETypeImageJPEG]`, so it still functions as expected.

PR #822, opened 2015-05-08
Feature/layerkit 0.13.0

kmahal:
This PR updates Atlas with the changes made in `LayerKit` v 0.13.0 that introduces the `LYRActor` object.  This object replaces the `LYRMessage` property `sentByUserID`.  `LYRActor` has properties `userID` and `name`, which allows for system messages in conversations.


chipxsd:
looks great

PR #825, opened 2015-05-13
Updated right button action to allow default message override

kmahal:


Code/Controllers/ATLConversationViewController.m
```diff
@@ -502,15 +502,13 @@ - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTap
     if (!self.conversation) {
         return;
     }
-    if (messageInputToolbar.mediaAttachments.count) {
-        NSOrderedSet *messages = [self messagesForMediaAttachments:messageInputToolbar.mediaAttachments];
+    NSOrderedSet *messages = [self messagesForMediaAttachments:messageInputToolbar.mediaAttachments];
```
kcoleman731: Looks good man. We are going to need to add documentation to the public data source method that the media attachments array can be nil...and what this means.

Code/Controllers/ATLConversationViewController.h
```diff
@@ -75,11 +75,11 @@
  @return An `NSOrderedSet` of `LYRMessage` objects. If `nil` is returned, the controller will fall back to default behavior. If an empty
  `NSOrderedSet` is returned, the controller will not send any messages.
  @discussion Called when a user taps the `SEND` button on an `ATLMessageInputToolbar`. The media attachments array supplied can contain
- any media type, such as text, images, GPS location information. Applications who wish to send `LYRMessage` objects with custom `LYRMessagePart`
```
kcoleman731: Let's update the old docs to be accurate....
```
 @discussion Called when a user taps the `leftAccessoryButton` on an `ATLMessageInputToolbar`.
```

Code/Controllers/ATLConversationViewController.h
```diff
@@ -75,11 +75,11 @@
...
- MIME types not supported by default by Atlas can do so by implementing this method. All `LYRMessage` objects returned will be immediately 
- sent into the current conversation for the controller. If implemented, applications should also register custom `UICollectionViewCell` classes 
- with the controller via a call to `registerClass:forMessageCellWithReuseIdentifier:`. They should also implement the optional data source method,
- `conversationViewController:reuseIdentifierForMessage:`.
+ any media type, such as text, images, GPS location information.  The media attachments will be empty when the location icon is tapped. Applications
```
kcoleman731: The media attachments array can also be empty, which indicates the `leftAccessoryButton` was tapped when it contained no content.

PR #832, opened 2015-05-18
Feature/public avatar property

thomasgpeng:
This is the avatar feature I discussed with @kmahal -- added public property `shouldDisplayAvatarItemForOneOtherParticipant` to `ATLConversationViewcontroller`.  Setting this property to YES will cause avatars to be shown even when there is only one other participant (besides the authenticated user) in the conversation.


kmahal:
Thanks @thomasgpeng.

PR #833, opened 2015-05-19
Make avatar initials nil when image is non-nil.

ccrazy88:


Code/Views/ATLAvatarImageView.m
```diff
@@ -97,8 +97,10 @@ - (void)setAvatarItem:(id<ATLAvatarItem>)avatarItem
 {
     if ([avatarItem avatarImageURL]) {
         [self loadAvatarImageWithURL:[avatarItem avatarImageURL]];
+        self.initialsLabel.text = nil;
```
kmahal: We'd like to keep the initials if the image hasn't yet loaded.  Add that validation and then it should be good to go.

kmahal:
@ccrazy88 I've annotated your PR.  Have you specifically seen the avatar initials on top of the image?  Or is this just a preference?

ccrazy88:
Hi! I have, but now that I think about it, the initials didn't always correspond to the user — for instance, the avatar and the message would belong to me, but the initial on top of the image would be someone else's. Might this be more of a cell reuse issue instead?

As far as the image URL validation goes, my bad, I can add that!

kmahal:
@ccrazy88 I've captured that change and a few others in: https://github.com/layerhq/Atlas-iOS/pull/834 so will close this PR.  I was making a change to the specific method and just roped it in.  Thank you for your input!

PR #834, opened 2015-05-20
Bug/apps 1579 avatar item choices

kmahal:
This makes the `ATLAvatarItem` non-optional so that developers will receive warnings when they have not implemented the specific getters.


Code/Views/ATLAvatarImageView.m
```diff
@@ -96,8 +96,10 @@ - (CGSize)intrinsicContentSize
 - (void)setAvatarItem:(id<ATLAvatarItem>)avatarItem
 {
     if ([avatarItem avatarImageURL]) {
+        self.initialsLabel = nil;
```
kcoleman731: Be consistent with your property access. Use either `.` or brackets...not both.

kmahal: @kcoleman731 the bracket notation was already used, and we the dot notation for direct property accessors in the class.  The bracket notation was used to signify the protocol properties, whereas dot notation is for the class's own properties.  But I can change it if you want more consistency.

PR #837, opened 2015-05-21
Attempt to fix avatar image reloading and reuse issues

ccrazy88:
- Added a "resetView" function to ATLAvatarImageView
- Use the function in "prepareForReuse" for all of the cells that it's in
- Change the setting to nil to the label's text rather than the label itself
- Add a comparison between the image URL fetched and the current URL when we're about to set it

### DIFF
--- a/Code/Views/ATLAvatarImageView.h
+++ b/Code/Views/ATLAvatarImageView.h
@@ -57,4 +57,9 @@ extern CGFloat const ATLAvatarImageDiameter;
  */
 @property (nonatomic) UIColor *imageViewBackgroundColor UI_APPEARANCE_SELECTOR;
 
+/**
+ @abstract Sets the avatar item, image view, and initial view to nil in preparation for reuse.
+ */
+- (void)resetView;
+
 @end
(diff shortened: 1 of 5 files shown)


ccrazy88:
Note: There are still some edge cases with the background image downloading (i.e. the same image is requested multiple times due to cell reuse so subsequent requests make the avatar flicker in and out even though the image is the same).

Code/Views/ATLAvatarImageView.h
```diff
@@ -57,4 +57,9 @@ extern CGFloat const ATLAvatarImageDiameter;
  */
 @property (nonatomic) UIColor *imageViewBackgroundColor UI_APPEARANCE_SELECTOR;
 
+/**
+ @abstract Sets the avatar item, image view, and initial view to nil in preparation for reuse.
+ */
+- (void)resetView;
```
blakewatters: Does this need to be in the public API?

ccrazy88: Since we use it in other places (ATLConversationTableViewCell, ATLMessageCollectionViewCell, ATLParticipantTableViewCell), I believe so... can you think of a more elegant way to accomplish this?

blakewatters: ah I see now. Sorry I did a very quick look earlier

PR #838, opened 2015-05-22
Fix reloadCellsForMessagesSentByParticipantWithIdentifier: crash

ccrazy88:
Uses collection view index paths rather than query controller index paths.


ccrazy88:
I think I have a better solution that will get rid of the animation. Hold on!

PR #840, opened 2015-05-26
Change public reloading functions to reconfigure cells.

ccrazy88:
Using the collection view to reload cells at index paths doesn't affect the header and the footer, unfortunately. Also, there is an animation that comes with the reloading method of the collection view that may or may not be desired. This is a replacement that uses code from the `configureCollectionViewElements` function in order to target specific cells.

It could probably be optimized further if it were based on a participant identifier but this seemed to work for us.


kmahal:
@ccrazy88 Thank you for this PR.  Merging.

PR #842, opened 2015-05-27
Expose -initWithLayerClient: method to allow for sub-classing while being able to call the super initialisation

Rupert-RR:
When creating a sub-class of ``ATLConversationListViewController`` there is no way to call ``super``'s initialiser from the sub-class, since it is not visible. This is a problem, especially in Swift, when making a sub-class which has properties which must be initialised while retaining ``ATLConversationListViewController``'s initialisation (please let me know if I'm not being clear).

### DIFF
--- a/Code/Controllers/ATLConversationListViewController.m
+++ b/Code/Controllers/ATLConversationListViewController.m
@@ -51,8 +51,9 @@ NSString *const ATLConversationTableViewAccessibilityIdentifier = @"Conversation
     return [[self alloc] initWithLayerClient:layerClient];
 }
 
-- (id)initWithLayerClient:(LYRClient *)layerClient
+- (instancetype)initWithLayerClient:(LYRClient *)layerClient
 {
+	NSAssert(layerClient, @"Layer Client cannot be nil");
     self = [super initWithStyle:UITableViewStylePlain];
     if (self)  {
         _layerClient = layerClient;
--- a/Code/Controllers/ATLConversationListViewController.h
+++ b/Code/Controllers/ATLConversationListViewController.h
@@ -145,6 +145,13 @@
  */
 + (instancetype)conversationListViewControllerWithLayerClient:(LYRClient *)layerClient;
 
+/**
+ @abstract Creates and returns a new conversation list initialized with a given `LYRClient` object.
+ @param layerClient The `LYRClient` object from which conversations will be fetched for display.
+ @return An `LYRConversationListViewController` object.
+ */
+- (instancetype)initWithLayerClient:(LYRClient *)layerClient;
+
 ///-------------------------------------------------------
 /// @name Configuring Layer Client, Delegate & Data Source
 ///-------------------------------------------------------
--- a/Code/Controllers/ATLConversationListViewController.h
+++ b/Code/Controllers/ATLConversationListViewController.h
@@ -150,7 +150,7 @@
  @param layerClient The `LYRClient` object from which conversations will be fetched for display.
  @return An `LYRConversationListViewController` object.
  */
-- (instancetype)initWithLayerClient:(LYRClient *)layerClient;
+- (instancetype)initWithLayerClient:(LYRClient *)layerClient NS_DESIGNATED_INITIALIZER;
 
 ///-------------------------------------------------------
 /// @name Configuring Layer Client, Delegate & Data Source
--- a/Code/Controllers/ATLConversationViewController.h
+++ b/Code/Controllers/ATLConversationViewController.h
@@ -180,6 +180,13 @@
 + (instancetype)conversationViewControllerWithLayerClient:(LYRClient *)layerClient;
 
 /**
+ @abstract Creates and returns a new `ATLConversationViewController` initialized with an `LYRClient` object.
+ @param layerClient The `LYRClient` object from which to retrieve the messages for display.
+ @return An `LYRConversationViewController` object.
+ */
+- (instancetype)initWithLayerClient:(LYRClient *)layerClient NS_DESIGNATED_INITIALIZER;
+
+/**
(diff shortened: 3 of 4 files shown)


Rupert-RR:
By the way, I did run the tests, but so many of them seem to be failing anyway (iOS 8.3 on simulator) that there didn't really seem to be any point.

kmahal:
@Rupert-RR is the method `+ (instancetype)conversationListViewControllerWithLayerClient:(LYRClient *)layerClient` not sufficient for initializing purposes?

Rupert-RR:
@kmahal Unfortunately not - this does not allow for sub-classes with custom initialisers (we're using Swift and have extra properties which must be initialised in the designated initialiser).
What is the logic behind not exposing this method anyway?

kmahal:
@Rupert-RR All Atlas specific properties are initialized in that `lyr_commonInit` method called in the private method.  For custom, non-Atlas properties, you can initialize them yourself in `viewDidLoad` or as public properties before presenting the controller on screen.  You can create your own subclass initializer which will call `+ (instancetype)conversationListViewControllerWithLayerClient:(LYRClient *)layerClient` and set your properties while returning the instance you desire.

Rupert-RR:
@kmahal  While I appreciate that this was not written for Swift, it does seem that Swift is the way forward.
If you were using this in Swift then, in a custom initialisation method, you would have to call one of the designated initialisers in the superclass (which means that if you have designated a particular initialiser then you can guarantee that it will be called).
If you are not using Swift then there are plenty of ObjC runtime ways to get around the hiding of the initialiser, so you are not preventing anything.
However, by not exposing an initialiser you are basically preventing any kind of meaningful sub-classing. In Swift there is an emphasis placed on using immutable (and preferably non optional) properties where possible (which avoids many programming errors and pitfalls). Such properties can only be initialised in an init method, which means that you are completely preventing their use in this class.
I understand that you have required initialisers that must be called, but surely in this case if the developer makes the effort to sub-class and override, the onus is on him to make sure he does so in a way which is compatible with the superclass.
The idea of not trusting the developer, and not giving him the choice, especially in an open-source library seems counter-productive. Our use of Atlas will depend on being able to make it work with our FRP paradigm, which at the moment means having to use a customised fork and could easily mean that we don't actually use Atlas other than for a bit of inspiration.

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -51,8 +51,9 @@ + (instancetype)conversationListViewControllerWithLayerClient:(LYRClient *)layer
     return [[self alloc] initWithLayerClient:layerClient];
 }
 
-- (id)initWithLayerClient:(LYRClient *)layerClient
+- (instancetype)initWithLayerClient:(LYRClient *)layerClient
 {
+	NSAssert(layerClient, @"Layer Client cannot be nil");
```
blakewatters: Fix indention

blakewatters:
Hi @Rupert-RR -

Kabir grabbed me to discuss this in depth. While we don’t currently have heavy investments in Swift we are certainly interested in ensuring that both LayerKit and Atlas play well as the language develops. We are extremely excited about the developments in Swift 2 and feel like the maturity is finally there for us to begin investing in Swift support.

In any case, I’ve reviewed your changes and find them perfectly suitable for merging. I understand the designated initializer patterns in Swift and will make sure the rest of the team is sensitive to these requirements as we extend and enhance the APIs.

blakewatters:
@Rupert-RR @kmahal I have merged this PR into master with 192514c955c98fbbf5896a2ff33ebddf014c1f7c

Cleanup changes include:
* Fixed mismatched spacing in one of the files.
* Tuned up the initializer language to better match and init method than a factory method.
* Rebased against master to eliminate merge commits.

Hope this helps.

kartikthapar:
> Unfortunately not - this does not allow for sub-classes with custom initialisers (we're using Swift and have extra properties which must be initialised in the designated initialiser).

I had similar issues too and the factory method was not very helpful. So I had to have my own initializers like:

    + (instancetype)conversationListViewControllerWithLayerClient:(LYRClient *)layerClient loadQuery:(LYRQuery *)query;

But in any case, this doesn't really affect the whole situation in terms of the factory API provided by Atlas. Simply call the factory API and set your properties in your own custom initializer.

+ (instancetype)conversationListViewControllerWithLayerClient:(LYRClient *)layerClient loadQuery:(LYRQuery *)query
{
    NSParameterAssert(query);
    
    MyCoversationListController *controller = [super conversationListViewControllerWithLayerClient:layerClient];
    
    // set the load query here
    controller.loadQuery = query;
    
    return controller;
}

So, I am not sure what the issue really is here. Exposing the `initWithLayerClient` API doesn't seem to influence the specified use case.

> What is the logic behind not exposing this method anyway?

I do agree here though — there is no reason for not exposing designated initializers.

Rupert-RR:
@blakewatters Thank you for taking the time to look this through.

@kartikthapar The problem with the factory API is specific to Swift - if you sub-class a class, and add your own immutable (let) properties to that class, those properties can then only be initialised in a custom init method in that class, which must then call a designated initialiser on its super-class (which is obviously not possible using the factory pattern).An immutable property cannot be set once its owner object has been initialised.

Rupert-RR:
@kartikthapar Thinking about it, an analgous problem in Objective C would be if you had private properties in your sub-class. In that case, once you have obtained your initial instance in your factory method, by calling the super-class factory method, you would be unable to set the private properties on it.

kartikthapar:
I don't know much about swift to comment on it. As far as Objective-C is concerned, nothing is really private. The `loadQuery` variable is in fact a property declared in the implementation file rather than the header file. :)

Issue #856, opened 2015-06-19
Display sender Avatar


ntnmrndn:
Hi, 

We would like to add this feature.
We are willing to implement it and submit a pull request.

Would you accept such feature ?
Any thoughts on how to implement it ?

kmahal:
Hi @ntnmrndn could you explain what you mean further?  Atlas already displays avatar items for conversations with more than 3 participants, and even a conversation with 2 participants if you set `@property (nonatomic) BOOL shouldDisplayAvatarItemForOneOtherParticipant`.  Do you mean for the authenticated user?

ntnmrndn:
Yes, for the current authenticated user.

kmahal:
@ntnmrndn got it.  If you wanted to submit a pull request, look at the layout functionality of the non-sender collection view cells that do display the avatar item.  That would be the best guideline to implementation.  In addition you would add a public property such as `shouldDisplayAuthenticatedUserAvatarItem` which would default to `NO` and have to be set to `YES` to display the avatar item.  Feel free to submit a PR for the feature as we would certainly accept it.  Thanks!

kmahal:
merged in https://github.com/layerhq/Atlas-iOS/pull/894

PR #861, opened 2015-06-23
Allow display of avatar for outgoign messages (current logged in user)

ntnmrndn:
See Feature Request: https://github.com/layerhq/Atlas-iOS/issues/856

### DIFF
--- a/Code/Controllers/ATLConversationViewController.h
+++ b/Code/Controllers/ATLConversationViewController.h
@@ -247,7 +247,14 @@
 /**
  @abstract A Boolean value that determines whether or not an avatar is shown if there is only one other participant in the conversation.
  @default `NO`.
+ Should be set before `[super viewDidLoad]` is called.
  */
 @property (nonatomic) BOOL shouldDisplayAvatarItemForOneOtherParticipant;
+
+/**
+ @abstract A Boolean value that determines whether or not an avatar is shown next to the outgoing messages
+ @default `NO`.
+ */
+@property (nonatomic) BOOL shouldDisplayAvatarItemForOutgoingMessages;
 
 @end
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -94,6 +94,7 @@ static NSString *const ATLDefaultPushAlertText = @"sent you a message.";
     _dateDisplayTimeInterval = 60*60;
     _marksMessagesAsRead = YES;
     _shouldDisplayAvatarItemForOneOtherParticipant = NO;
+    _shouldDisplayAvatarItemForOutgoingMessages = NO;
     _typingParticipantIDs = [NSMutableOrderedSet new];
     _sectionHeaders = [NSHashTable weakObjectsHashTable];
     _sectionFooters = [NSHashTable weakObjectsHashTable];
@@ -486,7 +487,7 @@ static NSString *const ATLDefaultPushAlertText = @"sent you a message.";
     if (!self.shouldDisplayAvatarItem) return NO;
    
     LYRMessage *message = [self.conversationDataSource messageAtCollectionViewIndexPath:indexPath];
-    if ([message.sender.userID isEqualToString:self.layerClient.authenticatedUserID]) {
+    if ([message.sender.userID isEqualToString:self.layerClient.authenticatedUserID] && !self.shouldDisplayAvatarItemForOutgoingMessages) {
         return NO;
     }
    
--- a/Code/Views/ATLIncomingMessageCollectionViewCell.m
+++ b/Code/Views/ATLIncomingMessageCollectionViewCell.m
@@ -20,18 +20,21 @@
 
 #import "ATLIncomingMessageCollectionViewCell.h"
 
-@interface ATLIncomingMessageCollectionViewCell ()
+extern CGFloat const ATLAvatarImageLeadPadding;
+extern CGFloat const ATLAvatarImageTailPadding;
 
-@property (nonatomic) NSLayoutConstraint *bubbleWithAvatarLeftConstraint;
-@property (nonatomic) NSLayoutConstraint *bubbleWithoutAvatarLeftConstraint;
+@interface ATLMessageCollectionViewCell ()
+
+- (void)configureLayoutConstraints;
+
+@property (nonatomic) NSLayoutConstraint *bubbleWithAvatarLeadConstraint;
+@property (nonatomic) NSLayoutConstraint *bubbleWithoutAvatarLeadConstraint;
 
 @end
 
 @implementation ATLIncomingMessageCollectionViewCell
 
 NSString *const ATLIncomingMessageCellIdentifier = @"ATLIncomingMessageCellIdentifier";
-CGFloat const ATLAvatarImageLeftPadding = 12.0f;
-CGFloat const ATLAvatarImageRightPadding = 7.0f;
 
 + (void)initialize
 {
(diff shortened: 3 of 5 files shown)


kmahal:
Hey @ntnmrndn I'll be reviewing your pull request.  One specific thing I noticed by glancing at is the formatting - please erase every new line you introduced, as well as the indentations that you changed.  I'll comment on implementation details in line.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -481,15 +483,22 @@ - (BOOL)shouldClusterMessageAtSection:(NSUInteger)section
...
-    if (!self.shouldDisplayAvatarItem) return NO;
-   
+    if (!self.shouldDisplayAvatarItem) {
+	return NO;
+    }
     LYRMessage *message = [self.conversationDataSource messageAtCollectionViewIndexPath:indexPath];
-    if ([message.sender.userID isEqualToString:self.layerClient.authenticatedUserID]) {
-        return NO;
+
+    if ([self senderIsCurrentAuthentificatedUserForMessage:message] && !self.shouldDisplayAvatarItemForOutgoingMessages) {
```
kmahal: create a C helper function in `ATLMessagingUtilities` instead of a method just used in this once instance.

Code/Views/ATLIncomingMessageCollectionViewCell.m
```diff
@@ -40,61 +37,15 @@ + (void)initialize
     proxy.messageLinkTextColor = ATLBlueColor();
 }
 
-- (id)initWithFrame:(CGRect)frame
```
kmahal: we want to keep the common initializers in each subclass still due to their own layout constraints.

Code/Views/ATLMessageCollectionViewCell_Subclass.h
```diff
@@ -0,0 +1,24 @@
+//
+//  ATLMessageCollectionViewCell_Subclass.h
```
kmahal: What is this class for?  What couldn't be implemented in an existing subclass?

Code/Views/ATLOutgoingMessageCollectionViewCell.m
```diff
@@ -32,28 +33,15 @@ + (void)initialize
...
 }
 
-- (id)initWithFrame:(CGRect)frame
-{
-    self = [super initWithFrame:frame];
-    if (self) {
-        [self lyr_outgoingCommonInit];
-    }
-    return self;
-}
```
kmahal: We want to keep common init and the layer standards.

Examples/ATLSampleConversationViewController.m
```diff
@@ -33,6 +33,9 @@ @implementation ATLSampleConversationViewController
 
 - (void)viewDidLoad
 {
+    // Setup avatars behaviors (default to NO)
+//    self.shouldDisplayAvatarItemForOneOtherParticipant = YES;
```
kmahal: remove commented out items.

kmahal:
I pulled it down and ran it - the behavior looks good.  However there a few questions on the implementation details that we'll have to fix.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -373,11 +375,11 @@ - (void)configureCell:(UICollectionViewCell<ATLMessagePresenting> *)cell forMess
 {
     [cell presentMessage:message];
     [cell shouldDisplayAvatarItem:self.shouldDisplayAvatarItem];
-
```
kmahal: an example of the extra new-lines introduced.

Code/Controllers/ATLConversationViewController.m
ntnmrndn: As it uses a reference to self, this would be a bit difficult. I rolled back to the previous code.
The exacte same snippet is used 5 time in the class. An other solution would be to let the method and refactor the 5 codes to call the method.

Code/Views/ATLMessageCollectionViewCell_Subclass.h
ntnmrndn: This is not a class, just a header to simulate protected access. I used it to access to some parent properties and methods from the subclass, without exposing it to others.

ntnmrndn:
I re pushed. 
About Code/Views/ATLMessageCollectionViewCell_Subclass.h, maybe renaming the header ATLMessageCollectionViewCell_Protected.h would make its usage clearer ?

kmahal:
@ntnmrndn let's just add the private interface to the top of the implementation file instead of creating the new header file completely.

ntnmrndn:
Pushed again.

ntnmrndn:
Is there anything else to add/edit ?

Code/Views/ATLOutgoingMessageCollectionViewCell.m
```diff
@@ -20,6 +20,19 @@
 
 #import "ATLOutgoingMessageCollectionViewCell.h"
 
+extern CGFloat const ATLAvatarImageLeadPadding;
+extern CGFloat const ATLAvatarImageTailPadding;
+
+@interface ATLMessageCollectionViewCell()
```
kmahal: Space before opening `(`

Code/Views/ATLOutgoingMessageCollectionViewCell.m
```diff
@@ -52,8 +65,15 @@ - (id)initWithCoder:(NSCoder *)aDecoder
 
 - (void)lyr_outgoingCommonInit
 {
-    self.avatarImageView.hidden = YES;
-    [self.contentView addConstraint:[NSLayoutConstraint constraintWithItem:self.bubbleView  attribute:NSLayoutAttributeRight relatedBy:NSLayoutRelationEqual toItem:self.contentView attribute:NSLayoutAttributeRight multiplier:1.0 constant:-ATLMessageCellHorizontalMargin]];
+    [self configureConstraintsForOutgoingMessage];
+}
+
+- (void)configureConstraintsForOutgoingMessage
```
kmahal: Few issues here that affect spacing, if you check all of the potential states.  Your layout constraint is setting the contentView in relation to other items.  It should be:
```objc
- (void)configureConstraintsForOutgoingMessage
{
    [self.contentView addConstraint:[NSLayoutConstraint constraintWithItem:self.avatarImageView attribute:NSLayoutAttributeRight relatedBy:NSLayoutRelationEqual toItem:self.contentView attribute:NSLayoutAttributeRight multiplier:1.0 constant:-ATLAvatarImageLeadPadding]];
    self.bubbleWithAvatarLeadConstraint = [NSLayoutConstraint constraintWithItem:self.avatarImageView attribute:NSLayoutAttributeLeft relatedBy:NSLayoutRelationEqual toItem: self.bubbleView attribute: NSLayoutAttributeRight multiplier:1.0 constant:ATLAvatarImageTailPadding];
    [self.contentView addConstraint:self.bubbleWithAvatarLeadConstraint];
    self.bubbleWithoutAvatarLeadConstraint = [NSLayoutConstraint constraintWithItem:self.bubbleView attribute:NSLayoutAttributeRight relatedBy:NSLayoutRelationEqual toItem: self.contentView attribute:NSLayoutAttributeRight multiplier:1.0 constant:-ATLMessageCellHorizontalMargin];
}
```
Since we are setting the right cell (outgoing), the constant used should also be negative for `ATLAvatarImageLeadPadding` and `ATLMessageCellHorizontalMargin` as shown above.

Code/Views/ATLOutgoingMessageCollectionViewCell.m
```diff
@@ -20,6 +20,19 @@
 
 #import "ATLOutgoingMessageCollectionViewCell.h"
 
+extern CGFloat const ATLAvatarImageLeadPadding;
+extern CGFloat const ATLAvatarImageTailPadding;
+
+@interface ATLMessageCollectionViewCell()
+
+- (void)configureLayoutConstraints;
```
kmahal: Why have this here?  `[super configureLayoutConstraints]` isn't being called anywhere.

Code/Views/ATLIncomingMessageCollectionViewCell.m
```diff
@@ -20,18 +20,21 @@
...
 
-@interface ATLIncomingMessageCollectionViewCell ()
+extern CGFloat const ATLAvatarImageLeadPadding;
+extern CGFloat const ATLAvatarImageTailPadding;
 
-@property (nonatomic) NSLayoutConstraint *bubbleWithAvatarLeftConstraint;
-@property (nonatomic) NSLayoutConstraint *bubbleWithoutAvatarLeftConstraint;
+@interface ATLMessageCollectionViewCell ()
+
+- (void)configureLayoutConstraints;
```
kmahal: Why have this here? [super configureLayoutConstraints] isn't being called anywhere.

Code/Controllers/ATLConversationViewController.h
```diff
@@ -247,7 +247,14 @@
...
  @default `NO`.
+ Should be set before `[super viewDidLoad]` is called.
  */
 @property (nonatomic) BOOL shouldDisplayAvatarItemForOneOtherParticipant;
 
+/**
+ @abstract A Boolean value that determines whether or not an avatar is shown next to the outgoing messages
+ @default `NO`.
+ */
+@property (nonatomic) BOOL shouldDisplayAvatarItemForOutgoingMessages;
```
kmahal: let's rename this to `shouldDisplayAvatarItemForAuthenticatedUser`

kmahal:
@ntnmrndn comments added.  In addition, there's an issue in the use case that `shouldDisplayAvatarItems` is true, but `shouldDisplayAvatarItemForOutgoingMessages` is not (any default conversation with more than 2 participants).  If you have the layout constraint improvements I mentioned above, the spacing is still off due to the `cell:shouldDisplayAvatarItem:` moved into the superclass `ATLMessageCollectionViewCell` instead of the subclasses:
```objc
- (void)configureCell:(UICollectionViewCell<ATLMessagePresenting> *)cell forMessage:(LYRMessage *)message indexPath:(NSIndexPath *)indexPath
{
    [cell presentMessage:message]; 
    [cell shouldDisplayAvatarItem:self.shouldDisplayAvatarItem];
    
    if ([self shouldDisplayAvatarItemAtIndexPath:indexPath]) {
        [cell updateWithSender:[self participantForIdentifier:message.sender.userID]];
    } else {
        [cell updateWithSender:nil];
    }
...
}
```
The line ` [cell shouldDisplayAvatarItem:self.shouldDisplayAvatarItem]` doesn't account for `shouldDisplayAvatarItemForOutgoingMessages`.  In the previous implementation the `ATLIncomingMessageCollectionViewCell` handled this properly and the `ATLOutgoingMessageCollectionViewCell` only had the constraint without the avatar item, so it never had this issue.  

Because both cells adhere to the `ATLMessagePresentingProtocol`, find new logic to handle this newly introduced property.

kmahal:
Something like this could work, but would need cleaner variable names:
```objc
    BOOL displayAvatarItem = self.shouldDisplayAvatarItem;
    if ([message.sender.userID isEqualToString:self.layerClient.authenticatedUserID]) {
        displayAvatarItem = shouldDisplayAvatarItem && self.shouldDisplayAvatarItemForOutgoingMessages;
    }
    [cell shouldDisplayAvatarItem:displayAvatarItem];
```
This will compensate for both states.

Code/Views/ATLIncomingMessageCollectionViewCell.m
```diff
@@ -20,18 +20,21 @@
...
+@property (nonatomic) NSLayoutConstraint *bubbleWithAvatarLeadConstraint;
+@property (nonatomic) NSLayoutConstraint *bubbleWithoutAvatarLeadConstraint;
 
 @end
 
 @implementation ATLIncomingMessageCollectionViewCell
 
 NSString *const ATLIncomingMessageCellIdentifier = @"ATLIncomingMessageCellIdentifier";
-CGFloat const ATLAvatarImageLeftPadding = 12.0f;
-CGFloat const ATLAvatarImageRightPadding = 7.0f;
```
blakewatters: All of these constants should be outside the `@implementation` block

Issue #818, opened 2015-05-07
Question — changing composer text view font size


kartikthapar:
Setting composer text view font size through either appearance proxy or via accessory methods is not correctly laying out text field contents. This is probably due to the use of dummy text view in for height calculations. Relayout inside `setFont:` doesn't seem to work.

Anyway to change the font for the text field instead of setting a custom value inside the `init` method in `ATLMessageComposeTextView`.

kmahal:
@kartikthapar if you know a fix feel free to submit a PR. If not we'll look into it but I can't guarantee a timeline.  Please provide a screenshot of the problem as well if possible.

kmahal:
Hey @kartikthapar any update?  If not we'll close this one out.

kartikthapar:
I haven't looked at this in a long time but I will spend some time on this today in the evening. As far as I remember this, it was insanely hard to get it right without manually setting the font inside the initializer. Nonetheless, I will report by tomorrow. Thanks.

kmahal:
I'll wait for your PR

PR #862, opened 2015-06-23
- composer text view font using UIAppearance

kartikthapar:
Understandably this is a bit of a hack but this is all I got!


kmahal:
Reviewing now, @kartikthapar.  Have you tested this implementation?  Does it solve your problem from https://github.com/layerhq/Atlas-iOS/issues/818?  Getting tests on this would be good too.

Code/Views/ATLMessageInputToolbar.m
```diff
@@ -240,6 +240,18 @@ - (void)setRightAccessoryButtonDisabledColor:(UIColor *)rightAccessoryButtonDisa
...
 
+- (void)setRightAccessoryButtonTitleLabelFont:(UIFont *)rightAccessoryButtonTitleLabelFont
+{
+    _rightAccessoryButtonTitleLabelFont = rightAccessoryButtonTitleLabelFont;
+    [self.rightAccessoryButton.titleLabel setFont:rightAccessoryButtonTitleLabelFont];
+}
+
+- (void)setTextInputFont:(UIFont *)textInputFont
+{
+    [self.textInputView setFont:textInputFont];
```
kmahal: you must set the property's value here too.

kmahal:
commented.  You must also set the default values in the default initialization method `+ (void)initialize`.

kartikthapar:
Ya its a pretty loose PR, was just trying to get my things working. I’ll have this updated with tests, etc in a while. Thanks.

— Kartik Thapar

kartikthapar:
There are issues with the current upstream; so I have revoked the PR. I will resubmit this in due time. Thanks.

PR #863, opened 2015-06-24
- option to set font for text in ATLMessageInputToolbar

kartikthapar:
The text view font is primarily set in the `ATLMessageComposeTextView` initializer. Therefore, the font is not set in the initializer for `ATLMessageInputToolbar` (minor hack).

Note: updated PR for #862. In reference to #818.


kmahal:
Thanks @kartikthapar.  Have you confirmed this PR fixes the layout constraint issue you mentioned in #818?  If you can provide a before and after screenshot that would be helpful.

kartikthapar:
Image below shows setting the font for compose view in the `ATLConversationViewController` subclass using:

    self.messageInputToolbar.textInputView.font = [UIFont fontWithName:@"Georgia" size:30];

old

Image below shows setting text input view font using the new `UI_APPEARANCE_SELECTOR` API:

    [[ATLMessageInputToolbar appearance] setTextInputView:[UIFont fontWithName:@"Georgia" size:24]];

new

The only potential issue here is the frame for the toolbar which depends on the frame of the text input view instance.

kartikthapar:
Also, want to point out that the accessory views are for some reason restricted to the bottom of the frame instead of sticking them to the center. I am not sure of the reasoning behind this but I didn't change it as there were specific comments about this requirement.

kmahal:
@kartikthapar The buttons must be kept at the bottom, similar to how iMessage functions.  I'll pull down this code and test the implementation.  Thanks for your work!

kartikthapar:
@kmahal Ok. That's understandble. Although, if your text size is bigger, they appear to be weird. In iMessage, its not really noticeable as apply uses the default font and size (-1, maybe). I'd rather given an option for that. Further, if the font size is small, they appear to be out of place as the toolbar sizes according to the size of the compose view (as mentioned before). I am not sure if resizing the accessory buttons is the way to go. I suggest leaving that workload to the developer.

kmahal:
@kartikthapar this PR looks good.  Before I merge, can you confirm that your initial question in: https://github.com/layerhq/Atlas-iOS/issues/818 was referring to the placeholder text alignment, and not entered text?  I tested setting the `messageInputToolbar.textInputView`'s font directly and the toolbar scales fine, though the placeholder is the only thing off.

kartikthapar:
That is true. When you start entering the text, it resizes the frame. So this whole PR is primarily about setting the frame as soon as the view is displayed by using one of the appearance properties. I am sorry wasn't very clear in the question — it was mainly about the placeholder alignment but it had other side effects as well.

kmahal:
@kartikthapar this is better solved by simply adding
```objc
    self.dummyTextView.font = self.textInputView.font;
```
in `layoutSubviews` before it sets its frame.  Let's go with this simpler approach.  Feel free to close this PR and submit a new one to keep the commit history cleaner.

kartikthapar:
That is strange as I thought I tried setting the font in the `layoutSubviews` method before. Nevertheless, I'll submit a new PR. Thanks.

PR #864, opened 2015-06-24
Ability to add custom cells for ATLAddresBarViewController

mailyan:
Added a new feature to add custom cells for ATLAddresBarViewController. The default cell is ATLParticipantTableViewCell. The feature is working and tested with default cells and custom cells.


Code/Controllers/ATLAddressBarViewController.m
```diff
@@ -71,6 +79,22 @@ - (void)viewDidLoad
     [self configureLayoutConstraintsForTableView];
 }
 
+-(void)viewWillAppear:(BOOL)animated
+{
+    [super viewWillAppear:animated];
+
```
kmahal: remove extra line

Code/Controllers/ATLAddressBarViewController.m
```diff
@@ -41,6 +46,11 @@ @implementation ATLAddressBarViewController
 - (void)loadView
 {
     self.view = [ATLAddressBarContainerView new];
+
```
kmahal: remove extra line

Code/Controllers/ATLAddressBarViewController.m
```diff
@@ -41,6 +46,11 @@ @implementation ATLAddressBarViewController
 - (void)loadView
 {
     self.view = [ATLAddressBarContainerView new];
+    
+    _sortType = ATLParticipantPickerSortTypeFirstName;
+    _displaysAvatarItem = NO;
+    _rowHeight = 56.f;
+    _cellClass = [ATLParticipantTableViewCell class];
```
kmahal: can't these be set in `viewDidLoad`?  The subclass can then call `[super viewDidLoad]` and overwrite these set properties in their own `viewDidLoad`

Code/Controllers/ATLAddressBarViewController.m
```diff
@@ -162,6 +221,15 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
     [self selectParticipant:participant];
 }
 
+#pragma mark - Cell Configuration
+
+- (void)configureCell:(UITableViewCell<ATLParticipantPresenting> *)participantCell atIndexPath:(NSIndexPath *)indexPath
+{
+    id<ATLParticipant> participant = self.participants[indexPath.row];
+
```
kmahal: remove extra line

kmahal:
@mailyan Solid PR.  I've annotated a few small changes.  We do require the default cell label text color to be `ATLBlueColor()`.  You can add a property that is a `UI_APPEARANCE_SELECTOR` that is defaulted on load but then can be customized.

mailyan:
@kmahal I have a question about `UI_APPEARANCE_SELECTOR` color property for `ATLAddresBarViewController`. If I give it a default value and change `ATLParticipantTableViewCell`'s `titleColor`'s appearance on load according to it then it will also override any previously set appearance for `ATLParticipantTableViewCell`. It will change the appearance for `ATLParticipantTableViewCell` in `ATLParticipantTableViewController` as well. Is this a normal behavior?

Code/Controllers/ATLAddressBarViewController.m
```diff
@@ -61,16 +66,34 @@ - (void)viewDidLoad
     self.tableView.translatesAutoresizingMaskIntoConstraints = NO;
     self.tableView.delegate = self;
     self.tableView.dataSource = self;
-    self.tableView.rowHeight = 56;
-    [self.tableView registerClass:[UITableViewCell class] forCellReuseIdentifier:ATLMParticpantCellIdentifier];
     self.tableView.keyboardDismissMode = UIScrollViewKeyboardDismissModeOnDrag;
     self.tableView.hidden = YES;
     [self.view addSubview:self.tableView];
     
+    _sortType = ATLParticipantPickerSortTypeFirstName;
```
kmahal: these 4 properties can be called through `self`

kmahal:
@mailyan good point, we don't want that.  Since there already exits
``objc
/**
 @abstract The color for the title label displayed in the cell. Default is black.
 */
@property (nonatomic) UIColor *titleColor UI_APPEARANCE_SELECTOR;
```
on `ATLParticipantTableViewCell`.  This can be solved by creating a new table view cell subclass for the `ATLAddressBarController`, or a better implementation is a subclass of `ATLParticipantTableViewCell` that can specify it's own selector value.

mailyan:
@kmahal I came up with another idea. We can customize  'ATLParticipantTableViewCell' appearance just for 'ATLAddressBarController' by using 'UIAppearance' 'appearanceWhenContainedIn:' method.

mailyan:
@kmahal  I set default value for `ATLAddressBarController` default cell to `ATLBlueColor()`. In `defaultCellTitleColor` property setter method I set title color for `ATLParticipantTableViewCell` only when it is contained in `ATLAddressBarController`. The only problem now is when setting `ATLParticipantTableViewCell` appearance title color it does not affect cell color in `ATLAddressBarController`. Is this a right behavior? 

I also tried subclassing `ATLParticipantTableViewCell`. The problem here is when changing super class appearance it changes subclass appearance as well and I think that kind of a behavior is not right .

mailyan:
Are there any updates on this?

kcoleman731:
@mailyan Not sure how this PR fell through the cracks for so long. If you are able to rebase. I'll take a look at merging it quickly.

mailyan:
@kcoleman731 I made a new PR. Please take a look #1188

kcoleman731:
Mered #1188

PR #866, opened 2015-06-25
Added ability to scale down image size.

timominous:


kmahal:
@timominous we already set the `thumbnailSize` for the thumbnail images, which the developer can specify already.  The developer can also choose to not download the full size image in their implementation.  Is this pull request still necessary?  I see it's for the full image, but that's why we have the thumbnail for compression size.

timominous:
@kmahal I just thought it would be better to have the option. The preview works great for the receiver as they are able to view the image almost instantly. However, it still means that the sender would have to send the full sized image before the receiver can see the preview. Also, I think users would prefer the larger image if a he/she chooses to implement a fullscreen preview (or save to library) module.

kcoleman731:
@timominous I'm doing a bit of repo clean up and am going to close this PR unless you have any objections.

timominous:
@kcoleman731 I have nothing against it. I'm closing it now.

PR #829, opened 2015-05-15
Feature/apps 1403 keyboard copying

kmahal:
This commit prevents the keyboard from closing when long pressing a cell to copy the content.

PR #870, opened 2015-06-26
Fix UIMenuController quirks.

ccrazy88:
- Fix issue where you can repeatedly long press on a cell
- Fix issue where the menu controller points up on top of a cell if it's too close to the top/bottom edges of the scroll view's content view minus insets
- Fix issue where the menu controller doesn't show up if the cell is larger than the scroll view's content view minus insets

There is one issue outstanding: if you have the keyboard up and you long press, because we need to become the first responder, the keyboard is dismissed.


kmahal:
Thank you @ccrazy88, excellent PR.  For the last issue, it's one that we've seen.  I went through every possible implementation and decided on: https://github.com/layerhq/Atlas-iOS/pull/829.  We have to override the first responder, but it still doesn't feel great.  The collection view delegate methods don't work unfortunately, so we'll wait to see if we need this before merging.

PR #871, opened 2015-06-26
Keep input text color, when adding media attachment.

sammcd:
Similarly to how the set font is kept, keep the same font color for the attributed string when inserting a media attachment.


sammcd:
There is more work to do here.

Issue #873, opened 2015-06-27
Crash in ATLConversationViewController


thomasgpeng:
Hey guys, I'm having some issues with new messages in ATLConversationViewController.  When a new message arrives, ATLConversationListViewController is being updated correctly with the unread message indicator.  I can navigate to ATLConversationViewController, read the new message, but when I go back to ATLConversationListViewController, the app crashes.  Not sure what's happening here, but the crash log seems to point to the disposal of animationQueue being the problem.  Here's the log:

0   libdispatch.dylib             	0x00000001973c11d8 _dispatch_xref_dispose + 16 (object.c:162)
1   libdispatch.dylib             	0x00000001973bda64 -[OS_dispatch_object _xref_dispose] + 20 (object.m:332)
2   AnonCrumb                     	0x000000010009269c -[ATLConversationViewController .cxx_destruct] + 32 (ATLConversationViewController.m:52)
3   libobjc.A.dylib               	0x0000000196d6ab18 object_cxxDestructFromClass(objc_object*, objc_class*) + 144 (objc-class.mm:397)
4   libobjc.A.dylib               	0x0000000196d77f34 objc_destructInstance + 88 (objc-runtime-new.mm:7042)
5   libobjc.A.dylib               	0x0000000196d77f8c object_dispose + 24 (objc-runtime-new.mm:7061)
6   UIKit                         	0x000000018a2cf918 -[UIResponder dealloc] + 112 (UIResponder.m:55)
7   UIKit                         	0x000000018a10d510 -[UIViewController dealloc] + 1900 (UIViewController.m:1799)
8   AnonCrumb                     	0x000000010007e734 -[ATLBaseConversationViewController dealloc] + 116 (ATLBaseConversationViewController.m:145)
9   AnonCrumb                     	0x000000010008a40c -[ATLConversationViewController dealloc] + 172 (ATLConversationViewController.m:166)
10  UIKit                         	0x000000018a0ac14c -[_UIViewControllerOneToOneTransitionContext dealloc] + 36 (UIViewControllerTransitioning.m:353)
11  UIKit                         	0x000000018a2e6adc -[_UIViewControllerKeyboardAnimationStyle dealloc] + 68 (_UIViewControllerKeyboardAnimationStyle.m:120)
12  UIKit                         	0x000000018a6855c0 -[UIInputViewPlacementTransition dealloc] + 172 (UIInputWindowController.m:390)
[...]

PR #880, opened 2015-07-02
Fix ATLConversationViewController.m

thomasgpeng:
FIx `queryControllerDidChangeContent` to prevent crashes from uninstantiated `self.collectionView`.  Refer to issue #873.


kmahal:
Hi @thomasgpeng I believe this merged PR also fixes your issue: https://github.com/layerhq/Atlas-iOS/pull/882/files.  It has the same check for the `collectionView` `nil`ness.  I'll close out this PR and the corresponding issue, but let me know if anything else happens.

PR #883, opened 2015-07-02
Feature/apps 1843 caching media attachments

kmahal:
This PR allows the `ATLConversationViewController` to cache any unsent media attachments when deallocating and inserts them again when the conversation is reshown.

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -190,11 +190,10 @@ static NSString *const ATLDefaultPushAlertText = @"sent you a message.";
     NSArray *mediaAttachments = [[[self class] sharedMediaAttachmentCache] objectForKey:self.conversation.identifier];
     for (int i = 0; i < mediaAttachments.count; i++) {
         ATLMediaAttachment *attachment = [mediaAttachments objectAtIndex:i];
-        BOOL shouldHaveLineBreak = i < mediaAttachments.count - 1;
+        BOOL shouldHaveLineBreak = (i < mediaAttachments.count - 1) || !(attachment.mediaMIMEType == ATLMIMETypeTextPlain);
         [self.messageInputToolbar insertMediaAttachment:attachment withEndLineBreak:shouldHaveLineBreak];
     }
     [[[self class] sharedMediaAttachmentCache] removeObjectForKey:self.conversation.identifier];
-    [self.messageInputToolbar configureRightAccessoryButtonState];
 }
 
 #pragma mark - Conversation Data Source Setup
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -30,6 +30,12 @@
 #import "ATLMediaAttachment.h"
 #import "ATLLocationManager.h"
 
+@interface ATLMessageInputToolbar ()
+
+- (void)configureRightAccessoryButtonState;
+
+@end
+
 @interface ATLConversationViewController () <UICollectionViewDataSource, UICollectionViewDelegate, ATLMessageInputToolbarDelegate, UIActionSheetDelegate, LYRQueryControllerDelegate, CLLocationManagerDelegate>
 
 @property (nonatomic) ATLConversationDataSource *conversationDataSource;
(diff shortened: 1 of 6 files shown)


chipxsd:
One thing to note here, `NSCache` might evict some (or all) objects from the cache under memory pressure, which in this case might look like you left the input bar with five images, but when coming back to the view there might be only one image left.

kmahal:
@chipxsd we store all of the attachments under one key.  any memory pressure would release the entire object.  I can make it an `NSMutableDictionary`, but I think the cache would suffice.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -160,10 +177,30 @@ - (void)viewDidAppear:(BOOL)animated
...
+{
+    [[[self class] sharedCache] setObject:self.messageInputToolbar.mediaAttachments forKey:self.conversation.identifier];
+}
+
+- (void)loadCachedMediaAttachments
+{
+    NSArray *mediaAttachments = [[[self class] sharedCache] objectForKey:self.conversation.identifier];
+    for (ATLMediaAttachment *attachment in mediaAttachments) {
+        if (attachment.mediaType == ATLMediaAttachmentTypeText) {
+            [self.messageInputToolbar.textInputView setText:attachment.textRepresentation];
```
kcoleman731: I think you will have to do something else here. What happens if I have text part, then image part, then another text part? The third test part will override everything else?

Code/Controllers/ATLConversationViewController.m
```diff
@@ -30,6 +30,12 @@
 #import "ATLMediaAttachment.h"
 #import "ATLLocationManager.h"
 
+@interface ATLMessageInputToolbar ()
+
```
kcoleman731: Not a big fan of the class extension here. Should we just make this method public? Why would we not?

Code/Controllers/ATLConversationViewController.m
```diff
@@ -146,6 +162,7 @@ - (void)viewWillAppear:(BOOL)animated
     if (!self.hasAppeared) {
         [self.collectionView layoutIfNeeded];
     }
+    if (!self.hasAppeared && [[[self class] sharedCache] objectForKey:self.conversation.identifier]) [self loadCachedMediaAttachments];
```
kcoleman731: expand the if statement

Code/Controllers/ATLConversationViewController.m
```diff
@@ -160,10 +177,30 @@ - (void)viewDidAppear:(BOOL)animated
 
 - (void)dealloc
 {
+    if (self.messageInputToolbar.mediaAttachments.count > 0) [self cacheMediaAttachments];
```
kcoleman731: Expand the if statment

Tests/ATLConversationViewControllerTest.m
```diff
@@ -97,6 +97,26 @@ - (void)testToVerifySentImageAppearsInConversationView
...
+    [baseViewController.navigationController dismissViewControllerAnimated:YES completion:nil];
+    [tester waitForAnimationsToFinish];
+
+    [self setupConversationViewController];
+    [baseViewController.navigationController pushViewController:self.viewController animated:YES];
+    toolBar = (ATLMessageInputToolbar *)[tester waitForViewWithAccessibilityLabel:@"Message Input Toolbar"];
+    expect(toolBar.textInputView.text).to.equal(@"Test");
+}
+
 #pragma mark - ATLConversationViewControllerDelegate
```
kcoleman731: I'd like to see a bit more test coverage here. You should test different media attachments and ensuring they send correctly.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -160,10 +177,30 @@ - (void)viewDidAppear:(BOOL)animated
...
+    for (ATLMediaAttachment *attachment in mediaAttachments) {
+        if (attachment.mediaType == ATLMediaAttachmentTypeText) {
+            [self.messageInputToolbar.textInputView setText:attachment.textRepresentation];
+        } else {
+            [self.messageInputToolbar insertMediaAttachment:attachment];
+        }
+    }
+    [[[self class] sharedCache] removeObjectForKey:self.conversation.identifier];
+    [self.messageInputToolbar configureRightAccessoryButtonState];
+}
```
kcoleman731: I think this `configureRightAccessortyButtonState` call  should be unnecessary? The bar should just update if you set its text or insert an attachement

Code/Controllers/ATLConversationViewController.m
```diff
@@ -58,6 +64,16 @@ @implementation ATLConversationViewController
 static NSString *const ATLDefaultPushAlertLocation = @"sent you a location.";
 static NSString *const ATLDefaultPushAlertText = @"sent you a message.";
 
++ (NSCache *)sharedCache
```
kcoleman731: lets give this a more descriptive name...`mediaAttachementCache`...or something.

Code/Controllers/ATLConversationViewController.m
kmahal: We totally can make it public, I'll do so.

Code/Controllers/ATLConversationViewController.m
kmahal: Unfortunately not.  That only occurs from `textView` delegate methods from user input, not from direct insertion as media attachments.

Code/Controllers/ATLConversationViewController.m
kmahal: I'll add in tests for that and test the implementation.

PR #886, opened 2015-07-06
Avatar Items

andrewcopp:
I know this violates your merge vs. rebase philosophy but I wanted to submit it and start getting feedback.

Our team needs to be able to display avatars for all participants. I tried to do it in a way that wouldn't affect existing users. Please let me know what you think.

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -496,8 +496,8 @@ static NSString *const ATLDefaultPushAlertText = @"sent you a message.";
         return NO;
     }
    
-    if ([self shouldClusterMessageAtSection:indexPath.section] && self.avatarItemDisplayFrequency != ATLAvatarItemDisplayFrequencyCluster) {
-        return NO;
+    if (![self shouldClusterMessageAtSection:indexPath.section] && self.avatarItemDisplayFrequency == ATLAvatarItemDisplayFrequencyCluster) {
+        return YES;
     }
     
     NSInteger lastQueryControllerRow = [self.conversationDataSource.queryController numberOfObjectsInSection:0] - 1;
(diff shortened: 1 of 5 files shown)


kmahal:
@andrewcopp overall the structure looks good.  If you see https://github.com/layerhq/Atlas-iOS/pull/861 they implement it similarly.  Because both subclasses can have avatar items, a lot of the common logic is moved to `ATLMessageCollectionViewCell` (the superclass), with just the specific layout constraints applied in the `lyr_CommonInit`.  The superclass can then change constraints in `- (void)shouldDisplayAvatarItem:(BOOL)shouldDisplayAvatarItem`.  Your `ATLAvatarItemDisplayFrequency` implementation looks solid.  Once this other PR gets refined and merged down (still has comments on it) then we can put your changes on top as well.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -372,7 +374,11 @@ - (void)scrollViewDidScrollToTop:(UIScrollView *)scrollView
 - (void)configureCell:(UICollectionViewCell<ATLMessagePresenting> *)cell forMessage:(LYRMessage *)message indexPath:(NSIndexPath *)indexPath
 {
     [cell presentMessage:message];
-    [cell shouldDisplayAvatarItem:self.shouldDisplayAvatarItem];
+    if ([message.sender.userID isEqual:self.layerClient.authenticatedUserID]) {
+        [cell shouldDisplayAvatarItem:self.shouldDisplayAvatarItemForUser];
```
kmahal: It's possible to encounter a case where `shouldDisplayAvatarItemForOthers` is `NO`, while `shouldDisplayAvatarItemForUser` is `YES`.  This would leave spaces on the right side of the outgoing cells for avatar items and not the left, which probably isn't desired behavior.  and because `shouldDisplayAvatarItemAtIndexPath:` returns `NO` for both cases `if (!self.shouldDisplayAvatarItemForOthers) return NO;`, then it would not show any avatar items either, which is not desired.

The other PR has an example snippet for this issue.

kmahal:
Closing this out due to the outgoing cell already merged PR and https://github.com/layerhq/Atlas-iOS/pull/896

PR #887, opened 2015-07-07
Expose Query controller

andrewcopp:


andrewcopp:
One major concern. The `LYRQueryController` is readonly when accessed through the `LYRConversationViewDataSource`. By exposing it through `LYRConversationViewController` it isn't protected and is subject to be modified. Not the end of the world but I want to make sure this doesn't break some intentional design.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -209,6 +210,7 @@ - (void)fetchLayerMessages
     
     self.conversationDataSource = [ATLConversationDataSource dataSourceWithLayerClient:self.layerClient query:query];
     self.conversationDataSource.queryController.delegate = self;
+    self.queryController = self.conversationDataSource.queryController;
```
kmahal: you can assign it to the public ivar, without needing the internal property as well.

kmahal:
@andrewcopp commented.  Because it's just a pointer to the object, even reassigning it wouldn't change the internal queryController from the data source.  Plus the public property you defined is readonly as well.  You can just assign that ivar directly without the internal property too.

Code/Controllers/ATLConversationViewController.m
kmahal: Actually our best practice is keeping the property  Ignore ^ comment.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -30,9 +30,10 @@
 #import "ATLMediaAttachment.h"
 #import "ATLLocationManager.h"
 
-@interface ATLConversationViewController () <UICollectionViewDataSource, UICollectionViewDelegate, ATLMessageInputToolbarDelegate, UIActionSheetDelegate, LYRQueryControllerDelegate, CLLocationManagerDelegate>
+@interface ATLConversationViewController () <UICollectionViewDataSource, UICollectionViewDelegate, ATLMessageInputToolbarDelegate, UIActionSheetDelegate, CLLocationManagerDelegate>
 
 @property (nonatomic) ATLConversationDataSource *conversationDataSource;
+@property (nonatomic) LYRQueryController *queryController;
```
kmahal: add `(nonatomic, readwrite)` even though it's readwrite by default, just to signify that we have a public readonly property.

PR #890, opened 2015-07-13
delegating datasource creation to queue off the main thread

kmahal:
This PR executes the `ATLConversationViewController` data source's querycontroller on a background thread to improve perceived performance.

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -46,6 +46,7 @@
 @property (nonatomic) BOOL shouldShareLocation;
 @property (nonatomic) BOOL canDisableAddressBar;
 @property (nonatomic) dispatch_queue_t animationQueue;
+@property (nonatomic) dispatch_queue_t dataSourceSetupQueue;
 
 @end
 
@@ -110,6 +111,7 @@ static NSInteger const ATLPhotoActionSheet = 1000;
     _sectionFooters = [NSHashTable weakObjectsHashTable];
     _objectChanges = [NSMutableArray new];
     _animationQueue = dispatch_queue_create("com.atlas.animationQueue", DISPATCH_QUEUE_SERIAL);
+    _dataSourceSetupQueue = dispatch_queue_create("com.atlas.dataSourceSetupQueue", DISPATCH_QUEUE_SERIAL);
 }
 
 - (void)loadView
@@ -238,11 +240,14 @@ static NSInteger const ATLPhotoActionSheet = 1000;
             @throw [NSException exceptionWithName:NSInvalidArgumentException reason:@"Data source must return an `LYRQuery` object." userInfo:nil];
         }
     }
-    
-    self.conversationDataSource = [ATLConversationDataSource dataSourceWithLayerClient:self.layerClient query:query];
-    self.conversationDataSource.queryController.delegate = self;
-    self.showingMoreMessagesIndicator = [self.conversationDataSource moreMessagesAvailable];
-    [self.collectionView reloadData];
+    dispatch_async(self.dataSourceSetupQueue, ^{
+        self.conversationDataSource = [ATLConversationDataSource dataSourceWithLayerClient:self.layerClient query:query];
+        dispatch_async(dispatch_get_main_queue(), ^{
+            self.conversationDataSource.queryController.delegate = self;
+            self.showingMoreMessagesIndicator = [self.conversationDataSource moreMessagesAvailable];
+            [self.collectionView reloadData];
+        });
+    });
 }
 
 #pragma mark - Conntroller Setup
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -240,12 +240,18 @@ static NSInteger const ATLPhotoActionSheet = 1000;
             @throw [NSException exceptionWithName:NSInvalidArgumentException reason:@"Data source must return an `LYRQuery` object." userInfo:nil];
         }
     }
+    self.conversationDataSource = [ATLConversationDataSource dataSourceWithLayerClient:self.layerClient query:query];
+
+    // Asynchronously executes the query controller to ensure the controller appears on screen as fast as possible without any main thread loading.
+    __weak typeof(self) weakSelf = self;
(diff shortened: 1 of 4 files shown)


Code/Controllers/ATLConversationViewController.m
```diff
@@ -238,11 +240,14 @@ - (void)fetchLayerMessages
             @throw [NSException exceptionWithName:NSInvalidArgumentException reason:@"Data source must return an `LYRQuery` object." userInfo:nil];
         }
     }
-    
-    self.conversationDataSource = [ATLConversationDataSource dataSourceWithLayerClient:self.layerClient query:query];
-    self.conversationDataSource.queryController.delegate = self;
-    self.showingMoreMessagesIndicator = [self.conversationDataSource moreMessagesAvailable];
-    [self.collectionView reloadData];
+    dispatch_async(self.dataSourceSetupQueue, ^{
```
chipxsd: add:
```objective-c
__weak typeof(self) weakSelf = self;
```
before the async call and use `weakSelf` instead of `self` inside the code block.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -238,11 +240,14 @@ - (void)fetchLayerMessages
...
+        self.conversationDataSource = [ATLConversationDataSource dataSourceWithLayerClient:self.layerClient query:query];
```
chipxsd: Don't set the member instances asynchronously.

chipxsd: I would suggest you move the `queryController.execute:` from the factory method or if you have to, do it asynchronously internally and add a `completionHandler:` to the factory method which you can execute on the main thread right after the `queryController.execute:`.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -238,11 +240,20 @@ - (void)fetchLayerMessages
...
-    self.showingMoreMessagesIndicator = [self.conversationDataSource moreMessagesAvailable];
-    [self.collectionView reloadData];
+
+    // Asynchronously executes the query controller to ensure the controller appears on screen as fast as possible without any main thread loading.
+    __weak typeof(self) weakSelf = self;
+    dispatch_async(self.dataSourceSetupQueue, ^{
+        NSError *error;
+        BOOL success = [weakSelf.conversationDataSource.queryController execute:&error];
+        if (!success) NSLog(@"LayerKit failed to execute query with error: %@", error);
+        dispatch_async(dispatch_get_main_queue(), ^{
```
blakewatters: You are still going through the callback in the event of a failure

Code/Controllers/ATLConversationViewController.m
```diff
@@ -238,11 +240,20 @@ - (void)fetchLayerMessages
...
-    self.conversationDataSource.queryController.delegate = self;
-    self.showingMoreMessagesIndicator = [self.conversationDataSource moreMessagesAvailable];
-    [self.collectionView reloadData];
+
+    // Asynchronously executes the query controller to ensure the controller appears on screen as fast as possible without any main thread loading.
+    __weak typeof(self) weakSelf = self;
+    dispatch_async(self.dataSourceSetupQueue, ^{
+        NSError *error;
+        BOOL success = [weakSelf.conversationDataSource.queryController execute:&error];
+        if (!success) NSLog(@"LayerKit failed to execute query with error: %@", error);
```
blakewatters: Expand this conditional

Code/Controllers/ATLConversationViewController.m
```diff
@@ -238,11 +240,20 @@ - (void)fetchLayerMessages
...
         }
     }
-    
     self.conversationDataSource = [ATLConversationDataSource dataSourceWithLayerClient:self.layerClient query:query];
-    self.conversationDataSource.queryController.delegate = self;
-    self.showingMoreMessagesIndicator = [self.conversationDataSource moreMessagesAvailable];
-    [self.collectionView reloadData];
+
+    // Asynchronously executes the query controller to ensure the controller appears on screen as fast as possible without any main thread loading.
+    __weak typeof(self) weakSelf = self;
```
blakewatters: You don’t actually need to do the weakSelf dance here because dispatch queues are single execution. You need to worry about constructs that copy their blocks and hold them (objects with completion handlers, `NSBlockOperation`, etc)

blakewatters: If you are going to do the `weakSelf` thing then you should take a strong reference inside the block. This ensures that you don’t complete the work halfway:

```objc
NSError *error = nil;
__strong typeof(weakSelf) strongSelf = weakSelf;
BOOL success = [strongSelf.conversationDataSource.queryController execute:&error];
if (!success) {
     NSLog(@"LayerKit failed to execute query with error: %@", error);
}
```

Code/Controllers/ATLConversationViewController.m
kmahal: @blakewatters in the previous implementation we log the error and continue with normal execution in either state.  I can definitely put the main thread operations `if (!success)`

Code/Models/ATLConversationDataSource.m
```diff
@@ -46,10 +46,6 @@ - (id)initWithLayerClient:(LYRClient *)layerClient query:(LYRQuery *)query
         _queryController = [layerClient queryControllerWithQuery:query];
         _queryController.updatableProperties = [NSSet setWithObjects:@"parts.transferStatus", @"recipientStatusByUserID", @"sentAt", nil];
         _queryController.paginationWindow = -numberOfMessagesToDisplay;
-        
-        NSError *error = nil;
-        BOOL success = [_queryController execute:&error];
-        if (!success) NSLog(@"LayerKit failed to execute query with error: %@", error);
```
blakewatters: Wouldn’t it be cleaner / more convenient to just add this support to LayerKit?

```objc
[_queryController executeWithCompletion:(void (^)(BOOL success, NSError *error) {
    if (success) {
        [self.collectionView reloadData];
    }
}];
```

blakewatters: We’d also want async query primitives on `LYRClient`: `executeQuery:completion:`, `countForQuery:completion:`

kmahal:
Closing this out due to a forthcoming LayerKit update.

PR #895, opened 2015-07-15
ATLConversationViewController delegate for cell customization

kmahal:
Added optional ATLConverstionViewController delegate method to pass delegate cells during initialization for further customization

### DIFF
--- a/Code/Controllers/ATLConversationViewController.h
+++ b/Code/Controllers/ATLConversationViewController.h
@@ -69,6 +69,16 @@
 - (CGFloat)conversationViewController:(ATLConversationViewController *)viewController heightForMessage:(LYRMessage *)message withCellWidth:(CGFloat)cellWidth;
 
 /**
+ @abstract Informs the delegate of a cell being configured for the specified message.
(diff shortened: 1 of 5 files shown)


Code/Controllers/ATLConversationViewController.h
```diff
@@ -69,6 +69,16 @@
 - (CGFloat)conversationViewController:(ATLConversationViewController *)viewController heightForMessage:(LYRMessage *)message withCellWidth:(CGFloat)cellWidth;
 
 /**
+ @abstract Informs the delegate of a cell being configured for the specified message.
+ @param viewController The `ATLConversationViewController` where the message cell will appear.
+ @param cell The `UICollectionViewCell` object that adheres to the `ATLMessagePresenting` protocol that will be displayed in the controller.
```
kcoleman731: conforms to a protocol

Podfile.lock
```diff
@@ -0,0 +1,48 @@
+PODS:
+  - Atlas (1.0.10):
+    - LayerKit (>= 0.14.2)
```
kcoleman731: Looks like you have an old LayerKit dependency here.

PR #899, opened 2015-07-21
Dismiss UIMenuController on any pan, and fix location data for cached location images

varkor:
This fixes an issue in which, if you open the context menu on a message bubble, then pan the scroll by first tapping on the message bubble (rather in the empty area to the side of the messages), then dragging, the context menu will not close.

Additionally, cached location images were not being loaded properly, having no location data being set (this affected some additional functionality that I've been working on).

### DIFF
--- a/Code/Views/ATLMessageBubbleView.m
+++ b/Code/Views/ATLMessageBubbleView.m
@@ -42,6 +42,7 @@ typedef NS_ENUM(NSInteger, ATLBubbleViewContentType) {
 @property (nonatomic) UIView *longPressMask;
 @property (nonatomic) CLLocationCoordinate2D locationShown;
 @property (nonatomic) UITapGestureRecognizer *tapGestureRecognizer;
+@property (nonatomic) UIPanGestureRecognizer *panGestureRecognizer;
 @property (nonatomic) NSURL *tappedURL;
 @property (nonatomic) NSLayoutConstraint *imageWidthConstraint;
 @property (nonatomic) MKMapSnapshotter *snapshotter;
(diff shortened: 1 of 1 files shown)


kmahal:
@varkor I'm not able to reproduce these issues.  Can you create a gif that shows the use case?  I long pressed a bubble to show the `UIMenuController`.  Panning outside the cell closes it on the first touch as expected.  The location images are working as expected as well so a further detailed explanation will be required to investigate.

varkor:
I'm sorry, I should have explained the situation more clearly to begin with. The difference in behaviour is when the pan action is initiated by dragging on the message bubble that the menu originated from. By dragging from the message bubble, one can displace the bubble from its menu entirely, which looks odd, and it not what users expect (see Apple's Messages app to see how they dismiss the menu upon any pan gesture, regardless of where it originates).

Current panning behaviour (focusing on the second pan):
original-panning

Adapted panning behaviour (in the pull request):
adapted-panning

The location issue is not noticeable to users at the moment. However, in the future it could well cause problems—if, for example, an option was added to open locations in Apple Maps. Maps that are loaded from the cache are treated simply as images, containing no position data. Thus, if you try to use the location's position data, you will be given an invalid `kCLLocationCoordinate2DInvalid` value. This is fixing a bug that won't affect users now, but could well rear its head in the future.

ccrazy88:
Thanks for fixing this! I noticed it a while ago but haven't had the time to whip up a PR—glad to see someone beat me to it :).

kmahal:
Thanks @varkor!  Much appreciated.

PR #900, opened 2015-07-21
Don't call sendLocationMessage if not necessary

andrewcopp:
Functionality and tests for not calling `sendLocationMessage` which would then results in requesting permissions.

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -565,7 +565,7 @@ static NSInteger const ATLPhotoActionSheet = 1000;
         return;
     }
     NSOrderedSet *messages = [self messagesForMediaAttachments:messageInputToolbar.mediaAttachments];
-    if (messages.count == 0) {
+    if (messages.count == 0 && messages) {
         [self sendLocationMessage];
     } else {
         for (LYRMessage *message in messages) {
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -565,7 +565,7 @@ static NSInteger const ATLPhotoActionSheet = 1000;
         return;
     }
     NSOrderedSet *messages = [self messagesForMediaAttachments:messageInputToolbar.mediaAttachments];
-    if (messages.count == 0 && messages) {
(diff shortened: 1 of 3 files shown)


Tests/ATLMessageInputBarTest.m
```diff
@@ -260,6 +260,25 @@ - (void)testToVerifySending5Photos
     [delegateMock verify];
 }
 
+- (void)testToVerifySendingWhitespaceDoesNotSendLocation
```
kmahal: The behavior change is actually in the `ATLConversationViewController`, not the message input toolbar.  Let's move the test there.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -565,7 +565,7 @@ - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTap
         return;
     }
     NSOrderedSet *messages = [self messagesForMediaAttachments:messageInputToolbar.mediaAttachments];
-    if (messages.count == 0) {
+    if (messages.count == 0 && messages) {
```
kmahal: Tested out the behavior and it should be `    if (messages.count == 0 && self.messageInputToolbar.displaysRightAccessoryImage) ` since this unfortunately won't let Atlas send location messages even if they want to.

kmahal: This actually won't let the whitespace problem be fixed.  One second.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -565,7 +565,7 @@ - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTap
         return;
     }
     NSOrderedSet *messages = [self messagesForMediaAttachments:messageInputToolbar.mediaAttachments];
-    if (messages.count == 0) {
+    if (messages.count == 0 && self.messageInputToolbar.displaysRightAccessoryImage) {
```
kmahal: should be `    if (messages.count == 0 && self.messageInputToolbar.textInputView.text.length == 0) ` since ^ would still occur in white space or in no space clicks.

PR #908, opened 2015-07-29
Feature/apps 1944 avatar image flicker

kcoleman731:
* Fixed issue that caused avatar image views to flicker when loading images.


Code/Views/ATLAvatarImageView.m
```diff
@@ -72,6 +74,7 @@ - (void)lyr_commonInit
     _initialsColor = [UIColor blackColor];
     _avatarImageViewDiameter = 27;
     
+    self.URLSession = [NSURLSession sharedSession];
```
chipxsd: If it's a shared session, do you really need a member var for it? Either use ephemeral session and have _it_ as a member variable, or use the singleton directly where needed.

chipxsd: Does the `ATLAvatarImageView` instance live inside of a cell or any other view that's reusable?
* If **yes**, how do you guarantee that this avatar view instance won't get recycled when an asynchronous image download is still in progress? Because you might expect that the _just-downloaded_ image (requested from before) would trump the new one.
* If **no** or if the reusable parent view instantiates a new `ATLAvatarImageView`, then how do you guarantee that instance is still alive when download completes? (null pointer reference assertion)

Code/Views/ATLAvatarImageView.m
```diff
@@ -93,10 +96,12 @@ - (CGSize)intrinsicContentSize
     return CGSizeMake(self.avatarImageViewDiameter, self.avatarImageViewDiameter);
 }
 
-- (void)resetView {
+- (void)resetView
+{
     self.avatarItem = nil;
     self.image = nil;
     self.initialsLabel.text = nil;
+    [self.downloadTask cancel];
```
blakewatters: You should also do this during dealloc

Code/Views/ATLAvatarImageView.m
kcoleman731: The answer is Yes, and that is the point of the PR. Whenever a cell is reused, the `resetView` method is called by the cell in which the avatar image view will reside. This will cancel any inflight request the image view possess. This all happens before the next request is made.

Code/Views/ATLAvatarImageView.m
kcoleman731: Done

PR #915, opened 2015-08-06
Image and gif loading now occurs on a non main thread, and then is bought back to the main thread for the image view update

kmahal:
This PR loads images and gifs on background threads when converting `NSdata` to `UImage` format, and then is set to the imageview on the main thread.  This will ensure smoother scrolling as there is no main thread delay as the data is converted.

### DIFF
--- a/ATLMessageCollectionViewCellTests.m
+++ b/ATLMessageCollectionViewCellTests.m
@@ -46,7 +46,7 @@
     [super tearDown];
 }
 
-- (void)testThatAsynchronousImageAndGifLoadingDoesNotUpdateReusedCells
+- (void)testThatAsynchronousGifLoadingDoesNotUpdateReusedCells
 {
     ATLMessageCollectionViewCell *cell = [[ATLMessageCollectionViewCell alloc] initWithFrame:CGRectZero];
     ATLMessageBubbleView *bubbleView = cell.bubbleView;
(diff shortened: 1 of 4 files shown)


chipxsd:
Can you add a kif test (no LayerKit in the mix) that verifies what happens when a giant gif gets asynchronously decoded and the cell is out of the view's scope?

Add a large .gif resource file (of your choice) to the UI test suite and create 20 text messages and a single message with that big gif content. Make the view start decoding the gif, then scroll back to the top, and using *OCMock* reject any `[self.bubbleView updateWithImage:displayingImage width:size.width];` calls.

chipxsd:
After a discussion with @kmahal, we decided that a KIF test doesn't make sense. But we can completely verify the behavior with a unit test.

Here's how:
- [ ] Create an instance of `ATLMessageCollectionViewCell` (`initWithFrame:yadayadayada:`...).
- [ ] Create a partial mock (`id partiallyMockedCell = OCMPartialMock(...)`) from that ^^^ instance and reject any calls to `updateWithImage:width:`.
- [ ] Create two `LYRMessageMock` objects
    - [ ] One containing an animated gif (let's call it `messageWithGif`).
    - [ ] One containing a text (let's call this one `messageWithText`).
- [ ] Set the `partiallyMockedCell.message = messageWithGif;`
- [ ] Call `configureBubbleViewForGIFContent` on the `partiallyMockedCell`.
- [ ] Now set the `partiallyMockedCell.message = messageWithText;`
- [ ] And call the `configureBubbleViewForGIFContent` again.
- [ ] Verify the behavior with `OCMVerifyAll(partiallyMockedCell)`.

In case the GIF gets decoded to fast for the test, use the same stalling technique you used in the previous attempt (blocking the messagePart.data with a semaphore).

chipxsd:
aww, you didn't check the checkboxes I prepared for you 😔

PR #916, opened 2015-08-07
Fix CollectionView insets in ATLBaseConversationViewController on iOS 9

kfweber:
- UIKeyboardWillShow notification was not being called, replaced and consolidated both UIKeyboardWillShow and UIKeyboardWillHide with UIKeyboardWillChangeFrame.

- CollectionView inset was not being set correctly from viewWillAppear, moved inset updates to viewDidLayoutSubviews.

### DIFF
--- a/Code/Controllers/ATLBaseConversationViewController.m
+++ b/Code/Controllers/ATLBaseConversationViewController.m
@@ -212,17 +212,13 @@ static CGFloat const ATLMaxScrollDistanceFromBottom = 150;
 
 #pragma mark - Notification Handlers
 
-- (void)keyboardWillShow:(NSNotification *)notification
-{
-    [self configureWithKeyboardNotification:notification];
-}
-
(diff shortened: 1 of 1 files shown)


kmahal:
@kfweber can you explain what you mean that `UIKeyboardWillShow` and `UIKeyboardWillHide` notifications aren't being called?  While testing the application both are being called and responded to correctly.

kfweber:
@kmahal We integrated Atlas into our project and found that those notifications were never being sent.   Only when we changed the notification to <code>UIKeyboardWillChangeFrame</code> was the selector called.  Our project was not removing the observer before posting, or had any other conflicts, so we thought it may be an iOS 9 issue.  If it's working for you, then I have no idea what the issue is or was.

kmahal:
Hey @kfweber I tested it again, and it looks to be working still.  I'm going to close this out, but feel free to open it up again if something else pops up.

PR #918, opened 2015-08-10
Feature/apps 1968 menu controller customization

kmahal:
This pull request allows developers to customize `UIMenuController` `UIMenuItem`s displayed on the `ATLMessageCollectionViewCell`.  By implementing `conversationViewController:configureCell:forMessage:`, developers can set `menuControllerActions` property of the `ATLBubbleView` with their own `UIMenuItem`s.  The selectors the developer creates passes the `UIMenuController` as the `(id)sender`, and using `ATLMessageFromATLMessageCollectionViewForMenuController` they can get the `LYRMessage` from the menu controller action.

### DIFF
--- a/Code/Utilities/ATLMessagingUtilities.m
+++ b/Code/Utilities/ATLMessagingUtilities.m
@@ -279,7 +279,7 @@ NSArray *ATLTextCheckingResultsForText(NSString *text, NSTextCheckingType linkTy
 
 #pragma mark - ATLConversationViewController menu controller methods
 
-LYRMessage *ATLMessageFromATLMessageCollectionViewCellForMenuController(UICollectionView *collectionview, UIMenuController *menuController)
+LYRMessage *ATLMessageFromATLMessageCollectionViewForMenuController(UICollectionView *collectionview, UIMenuController *menuController)
 {
     if (collectionview && menuController && [menuController isKindOfClass:[UIMenuController class]] && [collectionview isKindOfClass:[UICollectionView class]]) {
         CGRect frame = menuController.menuFrame;
--- a/Code/Utilities/ATLMessagingUtilities.m
+++ b/Code/Utilities/ATLMessagingUtilities.m
@@ -281,7 +281,7 @@ NSArray *ATLTextCheckingResultsForText(NSString *text, NSTextCheckingType linkTy
 
 LYRMessage *ATLMessageFromATLMessageCollectionViewCellForMenuController(UICollectionView *collectionview, UIMenuController *menuController)
 {
-    if ([menuController isKindOfClass:[UIMenuController class]]) {
+    if (collectionview && menuController && [menuController isKindOfClass:[UIMenuController class]] && [collectionview isKindOfClass:[UICollectionView class]]) {
         CGRect frame = menuController.menuFrame;
         CGPoint point = frame.origin;
         NSIndexPath *indexPath = [collectionview indexPathForItemAtPoint:point];
--- a/Code/Utilities/ATLMessagingUtilities.m
+++ b/Code/Utilities/ATLMessagingUtilities.m
@@ -276,18 +276,3 @@ NSArray *ATLTextCheckingResultsForText(NSString *text, NSTextCheckingType linkTy
     if (error) return nil;
     return [detector matchesInString:text options:kNilOptions range:NSMakeRange(0, text.length)];
 }
-
-#pragma mark - ATLConversationViewController menu controller methods
-
-LYRMessage *ATLMessageFromATLMessageCollectionViewForMenuController(UICollectionView *collectionview, UIMenuController *menuController)
-{
-    if (collectionview && menuController && [menuController isKindOfClass:[UIMenuController class]] && [collectionview isKindOfClass:[UICollectionView class]]) {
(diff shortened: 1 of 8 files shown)


Code/Utilities/ATLMessagingUtilities.m
```diff
@@ -275,3 +276,18 @@ void ATLLastPhotoTaken(void(^completionHandler)(UIImage *image, NSError *error))
     if (error) return nil;
     return [detector matchesInString:text options:kNilOptions range:NSMakeRange(0, text.length)];
 }
+
+#pragma mark - ATLConversationViewController menu controller methods
+
+LYRMessage *ATLMessageFromATLMessageCollectionViewForMenuController(UICollectionView *collectionview, UIMenuController *menuController)
```
kcoleman731: I'd put this logic in the controller itself. Passing an entire collection view object to a convenience is a bit strange.

Code/Views/ATLMessageBubbleView.h
```diff
@@ -91,4 +91,13 @@ extern NSString *const ATLUserDidTapPhoneNumberNotification;
...
 @property (nonatomic) NSTextCheckingType textCheckingTypes;
 
+/**
+ @abstract `UIMenuItem` objects that the application wants to display on long press, overriding the default copy functionality.
+ @discussion If set to nil or an empty array, this will deactivate all menu controller actions.  This property can be set in the
+ `ATLConversationViewController` subclass by implementing the delegate `conversationViewController:configureCell:forMessage` and
+ overriding this property's default value.
+ @default `UIMenuItem` copy action.
+ */
+@property (nonatomic) NSArray *menuControllerActions;
```
kcoleman731: is there a use case for different actions for different MIMEtypes? If so perhaps this should be a data source method.

kcoleman731: On second thought, I'm reading your comment for this property. Why is it even necessary?

Code/Views/ATLMessageBubbleView.m
```diff
@@ -228,6 +231,16 @@ - (void)applyImageWidthConstraint:(BOOL)applyImageWidthConstraint
     }
 }
 
+- (void)setMenuControllerActions:(NSArray *)menuControllerActions
+{
+    for (id object in menuControllerActions) {
```
kcoleman731: `for (UIMenuItem in`

Code/Views/ATLMessageBubbleView.h
kmahal: @kcoleman731 This was implemented to allow more than 1 menu action item for `ATLMessageCollectionViewCell`s.  It shouldn't be a data source method, as not all cells require these options.  Previously the only `UIMenuItem` we had was `copy:`, but this opens it up so that way in `conversationViewController:configureCell:forMessage:` developers can add their own `UIMenuItem`s to the native cells.

kmahal: the developer can also specify menu items based on `MIMEType` if they so choose.  I reviewed this with Blake, as my 2 proposed implementations were: https://gist.github.com/kmahal/c290b0f8e9ac08b30812 and this one is the best, we think.

varkor:
How exactly may we set the menu items?
If we implement `conversationViewController:configureCell:forMessage` we're passed a `UICollectionViewCell`, but we need an `ATLMessageBubbleView` to set `menuControllerActions`.
How do we get the particular `ATLMessageBubbleView` using the delegate method?

kmahal:
@varkor it doesn't pass a `ATLMessageCollectionViewCell` to retain Atlas' full customization capabilities.  The default cells provided implement the `ATLMessagePresenting` protocol, which allows the developers to create their own cells for custom MIMETypes or even for default ones.  The collection view does not know (or care) about the specific cell class, just that it implements the protocol, thus the delegate method passes a cell adhering to the protocol.

You can check `if ([cell isMemberOfClass:[ATLMessageCollectionViewCell class]])` and then get the `bubbleView` property, and customize the menu controller items as you see fit.

varkor:
Ah, all right, that makes sense—thanks!

varkor:
Another quick question: why was the convenience method removed? If we wanted to get the MIME type of the message, and perhaps more specific information such as the location of a shared location, what would be the best way to go about it? Reimplement the same technique as was previously in the convenience method?

kmahal:
@varkor In Atlas we want to give developers the abilities to do things themselves, rather than having convenience methods for every possible use case (which could get out of hand).  If you want the exact location, look at how the cell renders it:
```objc
- (void)configureBubbleViewForLocationContent
{
    LYRMessagePart *messagePart = self.message.parts.firstObject;
    NSDictionary *dictionary = [NSJSONSerialization JSONObjectWithData:messagePart.data
                                                               options:NSJSONReadingAllowFragments
                                                                 error:nil];
    double lat = [dictionary[ATLLocationLatitudeKey] doubleValue];
    double lon = [dictionary[ATLLocationLongitudeKey] doubleValue];
    [self.bubbleView updateWithLocation:CLLocationCoordinate2DMake(lat, lon)];
    [self.bubbleView updateProgressIndicatorWithProgress:0.0 visible:NO animated:NO];
}
```

By getting the message part you can find the information directly.

PR #924, opened 2015-08-14
Fix constraint in Bubble View

garhiggins:


kmahal:
Thanks @garhiggins.  Merged.

PR #931, opened 2015-08-19
Fix disabled addressbar listing of participants

artonragsdale:
Use plural 'others' string for multiple participants
Use singular 'participant' for one participant
Calculate participants count for string correctly
Set disabled string after views have been laid out


kcoleman731:
@artonragsdale I'm doing a bit of repo cleanup and it looks like this PR slipped through the cracks. It looks reasonable to me however. If you'd like to rebase, I'll take a look at getting it merged quickly. Thanks!

PR #933, opened 2015-08-19
Added feature to choose if receipts are displayed

kmahal:


Code/Controllers/ATLConversationViewController.m
```diff
@@ -495,6 +496,7 @@ - (BOOL)shouldDisplaySenderLabelForSection:(NSUInteger)section
 
 - (BOOL)shouldDisplayReadReceiptForSection:(NSUInteger)section
 {
+    if (!self.shouldDisplayReceipts) return NO;
```
chipxsd: For the sake of breakpoints:
```objective-c
if (!self.shouldDisplayReceipts) {
    return NO;
}
```

PR #935, opened 2015-08-20
Fixes bug which could cause images to paste with wrong orientation

kcoleman731:
Issue was caused by setting a `UIPasteboard` object's image property when copying an image. The UIPasteBoard would drop the image's `imageOrientation` property when extracting upon paste. 

This is remedied by serializing the image data and using `setData:forPasteboardType:` to copy the image to the pasteboard.

### DIFF
--- a/Code/Views/ATLMessageInputToolbar.m
+++ b/Code/Views/ATLMessageInputToolbar.m
@@ -173,7 +173,6 @@ static CGFloat const ATLButtonHeight = 28.0f;
                                                                                   metadata:nil
                                                                              thumbnailSize:ATLDefaultThumbnailSize];
         [self insertMediaAttachment:mediaAttachment withEndLineBreak:YES];
-
     }
 }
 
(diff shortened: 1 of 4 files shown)


Code/Views/ATLMessageInputToolbar.m
```diff
@@ -166,15 +166,14 @@ - (void)layoutSubviews
...
-        }
-        return;
+    NSData *imageData = [[UIPasteboard generalPasteboard] dataForPasteboardType:ATLPasteboardImageKey];
+    if (imageData) {
+        UIImage *image = [UIImage imageWithData:imageData];
+        ATLMediaAttachment *mediaAttachment = [ATLMediaAttachment mediaAttachmentWithImage:image
+                                                                                  metadata:nil
+                                                                             thumbnailSize:ATLDefaultThumbnailSize];
+        [self insertMediaAttachment:mediaAttachment withEndLineBreak:YES];
+
```
kmahal: extra white space line

PR #946, opened 2015-09-01
Feature/ATLMessageInputToolbar customizations

ricardorover:
This will allow:
- instantiate a custom ATLMessageInputToolbar in ATLBaseConversationViewController extended object
- change ATLMessageInputToolbar verticalMargin property
- implement custom right accessory behavior by overriding method "configureRightAccessoryButtonState".

### DIFF
--- a/Code/Controllers/ATLBaseConversationViewController.h
+++ b/Code/Controllers/ATLBaseConversationViewController.h
@@ -68,6 +68,11 @@
  */
 @property (nonatomic) BOOL displaysAddressBar;
 
+/**
+ @abstract Initializes the input accessory view of the ATLBaseConversationViewController, which by default is an instance of ATLMessageInputToolbar. Override this method to return a subclass of ATLMessageInputToolbar.
+ */
+- (ATLMessageInputToolbar *)initializeMessageInputToolbar;
+
 ///-------------------------------------
 /// @name Managing Scrolling
 ///-------------------------------------
--- a/Code/Views/ATLMessageInputToolbar.h
+++ b/Code/Views/ATLMessageInputToolbar.h
@@ -149,7 +149,7 @@ extern NSString *const ATLMessageInputToolbarAccessibilityLabel;
   @abstract The margin on top and bottom of the textInputView.
   @default 7.0f.
   */
-@property (nonatomic) CGFloat verticalMargin;
+@property (nonatomic) CGFloat verticalMargin UI_APPEARANCE_SELECTOR;
 
 /**
  @abstract The delegate object for the view.
@@ -180,5 +180,11 @@ extern NSString *const ATLMessageInputToolbarAccessibilityLabel;
  to support UISplitViewController usage.  This property should only be set when subclassing `ATLMessageInputToolbar`.
  */
 @property (nonatomic, weak) UIViewController *containerViewController;
+
+/**
+ @abstract Configures rightAccessoryButton on setup, when textInputView value changes and when media is attached. Override this method to customize right button behavior and appearance.
+ @default Configures button with atl.messagetoolbar.send.key label if textInputView has text value or image attached, shows button with rightAccessoryImage otherwise.
+ */
+- (void)configureRightAccessoryButtonState;
 
 @end
--- a/Code/Controllers/ATLBaseConversationViewController.m
+++ b/Code/Controllers/ATLBaseConversationViewController.m
@@ -75,7 +75,7 @@ static CGFloat const ATLMaxScrollDistanceFromBottom = 150;
     [super viewDidLoad];
     
     // Add message input tool bar
-    self.messageInputToolbar = [ATLMessageInputToolbar new];
+    self.messageInputToolbar = [self initializeMessageInputToolbar];
     // Fixes an ios9 bug that causes the background of the input accessory view to be black when being presented on screen.
     self.messageInputToolbar.translucent = NO;
     // An apparent system bug causes a view controller to not be deallocated
@@ -99,6 +99,11 @@ static CGFloat const ATLMaxScrollDistanceFromBottom = 150;
         [self configureAddressbarLayoutConstraints];
     }
     [self atl_baseRegisterForNotifications];
+}
+
+- (ATLMessageInputToolbar *)initializeMessageInputToolbar
+{
+    return [ATLMessageInputToolbar new];
 }
 
 - (void)viewWillAppear:(BOOL)animated
--- a/Code/Views/ATLMessageInputToolbar.m
+++ b/Code/Views/ATLMessageInputToolbar.m
@@ -46,7 +46,6 @@ NSString *const ATLMessageInputToolbarSendButton  = @"Message Input Toolbar Send
 // Compose View Margin Constants
 static CGFloat const ATLLeftButtonHorizontalMargin = 6.0f;
 static CGFloat const ATLRightButtonHorizontalMargin = 4.0f;
-static CGFloat const ATLVerticalMargin = 7.0f;
 
 // Compose View Button Constants
 static CGFloat const ATLLeftAccessoryButtonWidth = 40.0f;
@@ -60,6 +59,7 @@ static CGFloat const ATLButtonHeight = 28.0f;
     proxy.rightAccessoryButtonActiveColor = ATLBlueColor();
     proxy.rightAccessoryButtonDisabledColor = [UIColor grayColor];
     proxy.rightAccessoryButtonFont = [UIFont boldSystemFontOfSize:17];
+    proxy.verticalMargin = 7.0f;
 }
 
 - (id)init
@@ -90,8 +90,6 @@ static CGFloat const ATLButtonHeight = 28.0f;
         self.textInputView.layer.borderWidth = 0.5;
         self.textInputView.layer.cornerRadius = 5.0f;
         [self addSubview:self.textInputView];
-        
-        self.verticalMargin = ATLVerticalMargin;
         
         self.rightAccessoryButton = [[UIButton alloc] init];
         [self.rightAccessoryButton addTarget:self action:@selector(rightAccessoryButtonTapped) forControlEvents:UIControlEventTouchUpInside];


Code/Controllers/ATLBaseConversationViewController.h
```diff
@@ -68,6 +68,11 @@
  */
 @property (nonatomic) BOOL displaysAddressBar;
 
+/**
+ @abstract Initializes toolbar, allow override to use a custom ATLMessageInputToolbar.
+ */
+- (ATLMessageInputToolbar *)initializeMessageInputToolbar;
```
kmahal: This implementation logic is incorrect.  This does not change the input accessory view of the `ATLConversationViewController`, which is the actual desired behavior.  A subclass that implemented this logic would override the super class's behavior, effectively just returning an unused instance of your toolbar.

Code/Views/ATLMessageInputToolbar.h
```diff
@@ -154,9 +154,18 @@ extern NSString *const ATLMessageInputToolbarAccessibilityLabel;
 @property (nonatomic) NSUInteger maxNumberOfLines;
 
 /**
+ @abstract The margin on top and bottom of textInputView
+ @default 7.0f
+ */
+@property (nonatomic) NSUInteger verticalMargin;
```
kmahal: Adding properties like this can get out of hand - subclassing is the best approach here for customization to this level.

Code/Views/ATLMessageInputToolbar.h
```diff
@@ -154,9 +154,18 @@ extern NSString *const ATLMessageInputToolbarAccessibilityLabel;
...
+
+/**
  @abstract An array of all media attachments displayed in the text view.
  @discussion Any existing media attachments will be removed when the right accessory button is tapped.
  */
 @property (nonatomic, readonly) NSArray *mediaAttachments;
 
+
+- (void)configureRightAccessoryButtonState;
```
kmahal: Must add docs.  Additionally, properties set here and meant to be customized via `UI_APPEARANCE` selectors.

kmahal:
@ricardorover I've commented on your PR.  I've added several comments to it with implementation issues that must be resolved before we can proceed.

kmahal:
@ricardorover any follow up? If not I'll be closing out this PR.

Code/Controllers/ATLBaseConversationViewController.h
ricardorover: I did not understand your point. I've changed "self.messageInputToolbar = [ATLMessageInputToolbar new];" to "self.messageInputToolbar = [self initializeMessageInputToolbar];" to allow any ATLBaseConversationViewController subclass override "initializeMessageInputToolbar" method, thus initialising a custom ATLMessageInputToolbar. I can't see why the returned instance would be unused.

Code/Views/ATLMessageInputToolbar.h
ricardorover: I revisited this code, as I already have a ATLMessageInputToolbar subclass I've tried to do this change on the subclass only, doing the verticalMargin change in layoutSubviews, as ATLMessageInputToolbar does. It does not work, setting self.frame starts layoutSubviews all over again, and also seemed wrong. If there is some tutorial or a way to change the verticalMargin in a better way compared to the PR code, please make me know.

Code/Views/ATLMessageInputToolbar.h
ricardorover: Yes, I will add docs to configureRightAccessoryButtonState declaration.

Your comment about UI_APPEARANCE is related to property verticalMargin? If yes, I agree.

ricardorover:
@kmahal sorry the delay. I've replied with my questions.

Code/Controllers/ATLBaseConversationViewController.h
kmahal: If your subclass overrides `initializeMessageInputToolbar` with your own custom `ATLMessageInputToolbar` instance, and you set `self.messageInputToolbar = [self initializeMessageInputToolbar]`, this would not actually change the `inputAccessoryView` that the keyboard owns.  This would only change the pointer `self.messageInputToolbar`, and not change the UI.

If you want to allow customization, then make `ATLConversationView *view` a public property in `ATLBaseConversationViewController` to allow the developer to directly set the `inputAccessoryView` property of the view.

ricardorover: The `inputAcessoryView` that the keyboard owns is set by `self.view.inputAccessoryView = self.messageInputToolbar`, is that right? When a subclass overrides `initializeMessageInputToolbar`, it just changes `[ATLMessageInputToolbar new]` to anything the subclass wants (must be a subclass of `ATLMessageInputToolbar`). 

In `ATLBaseConversationViewController` at line 79, immediately after instantiating a `ATLMessageInputToolbar` the `inputAccessoryView` is set with the new created instance, so keyboard owns the right view (the view instantiated by `initializeMessageInputToolbar`).

Also, if I undo all `initializeMessageInputToolbar` related code, and just set `ATLConversationView *view` as a public property, the developer will be able to set `view.inputAccessoryView`, but all the `messageInputToolbar` references inside `ATLBaseConversationViewController` and `ATLConversationViewController` will hold a unused `ATLMessageInputToolbar` instance, and the developer will have to handle all behaviors that are already programmed in those 2 classes (anywhere that calls `self.messageInputToolbar`).

If you think this conversation is going nowhere please be free to close this PR, but I would like to understand how is the right way to implement this feature, and then fix this PR.

jeffdgr8: This logic appears sound to me. I've had to subclass the `ATLMessageInputToolbar` myself in order to keep it from covering the tab bar that contains the messaging view controller. Without a mechanism to override the instantiation of the message input toolbar, I have to duplicate much of the init code to reinitialize the message input toolbar after instantiating the subclass. This includes casting `self` to `id<ATLMessageInputToolbarDelegate>` in order to set the `ATLMessageInputToolbar` subclass' `inputToolBarDelegate` property, since `ATLConversationViewController`'s implementation of `ATLMessageInputToolbarDelegate` is hidden. It'd be much cleaner and avoid breaking if the view controller ever does anything differently to the message input toolbar during init if there was a way to explicitly instantiate a subclass of `ATLMessageInputToolbar`.

The other changes in this PR are unnecessary for my specific subclassing of `ATLMessageInputToolbar`, but the ability to override the instantiation of the message input toolbar by instantiating a subclass and have that used helps in my use case.

kmahal: @ricardorover Apologies, you are right about this, I misread the implementation.  Thanks for the PR.  I'll add a few more comments and we can get this merged down.

Code/Views/ATLMessageInputToolbar.h
kmahal: @ricardorover Can you explain the use case you'd like to override the default behavior that cannot be set otherwise?  Additionally lets get some docs on this.

Code/Controllers/ATLBaseConversationViewController.h
```diff
@@ -68,6 +68,11 @@
  */
 @property (nonatomic) BOOL displaysAddressBar;
 
+/**
+ @abstract Initializes toolbar, allow override to use a custom ATLMessageInputToolbar.
```
kmahal: Make this "Initializes the input accessory view of the `ATLBaseConversationViewController`, which by default is an instance of `ATLMessageInputToolbar`.  Override this method to return a subclass of `ATLMessageInputToolbar`."

kmahal:
Hi @ricardorover I've added a few comments.   The PR is no longer mergeable, so if you rebase on master too then we can get it in.  Thanks!

Code/Views/ATLMessageInputToolbar.h
ricardorover: Overriding configureRightAccessoryButtonState method now is possible, for example, change the "send location" behavior to "start record voice message" and change "Send" text to an icon with the same semantics (the requirements in my project).

ricardorover:
In the next few days I will do those adjusts. Thank you

DanielsCode:
looking forward to it .... great work! thanks

jeffdgr8:
Any update on when this PR can be ready for merging?

ricardorover:
Done! If some adjust is still required, please let me know.

wallerdev:
:+1: would like to see this merged

Code/Views/ATLMessageInputToolbar.h
```diff
@@ -181,4 +181,10 @@ extern NSString *const ATLMessageInputToolbarAccessibilityLabel;
  */
 @property (nonatomic, weak) UIViewController *containerViewController;
 
+/**
+ @abstract Configures rightAccessoryButton on setup, when textInputView value changes and when media is attached. Override this method to customize right button behavior and appearance.
+ @default Configures button with atl.messagetoolbar.send.key label if textInputView has text value or image attached, shows button with rightAccessoryImage otherwise.
+ */
+- (void)configureRightAccessoryButtonState;
```
kmahal: @ricardorover A few changes have been implemented in Atlas, including allowing the developer to choose a default right button text.  Additionally, the developer can choose to remove the current target and add their own selector to the right button action to override this behavior completely.  I don't think we need this here but happy to discuss.  Ideally we can open up another PR with this if you'd like to further comment so the message input toolbar changes can be merged promptly.

Code/Views/ATLMessageInputToolbar.h
```diff
@@ -149,7 +149,7 @@ extern NSString *const ATLMessageInputToolbarAccessibilityLabel;
   @abstract The margin on top and bottom of the textInputView.
   @default 7.0f.
   */
-@property (nonatomic) CGFloat verticalMargin;
+@property (nonatomic) CGFloat verticalMargin UI_APPEARANCE_SELECTOR;
```
kmahal: remove `UI_APPEARANCE_SELECTOR` tag.  The `vertical margin` property has already been merged so you can undo your changes and just leave the message input toolbar subclassing method.

ricardorover:
Hey @kmahal, I created a pull request from a feature branch (not from master like this one, sorry about that), with the code that allow use a custom ATLMessageInputToolbar: #1075
Also, I created another pull request for the custom right accessory button issue, #1076, to discuss the need and an implementation of this feature.

If that's all right I'm closing this pull request.

PR #950, opened 2015-09-02
Aligning Atlas with LayerKit v0.16.0 Changes

kmahal:


Code/Controllers/ATLConversationListViewController.m
```diff
@@ -229,9 +229,12 @@ - (void)setupConversationDataSource
         }
     }
     
-    self.queryController = [self.layerClient queryControllerWithQuery:query];
-    self.queryController.delegate = self;
     NSError *error;
+    self.queryController = [self.layerClient queryControllerWithQuery:query error:&error];
+    if (!self.queryController) {
+        NSLog(@"LayerKit failed to create a query controller with error: %@", error);
```
chipxsd: The error condition is not getting handled which makes the behavior after this line undefined.

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -468,9 +471,13 @@ - (BOOL)searchDisplayController:(UISearchDisplayController *)controller shouldRe
             LYRQuery *query = [LYRQuery queryWithQueryableClass:[LYRConversation class]];
             query.predicate = [LYRPredicate predicateWithProperty:@"participants" predicateOperator:LYRPredicateOperatorIsIn value:participantIdentifiers];
             query.sortDescriptors = @[[NSSortDescriptor sortDescriptorWithKey:@"lastMessage.receivedAt" ascending:NO]];
-            self.searchQueryController = [self.layerClient queryControllerWithQuery:query];
             
             NSError *error;
+            self.searchQueryController = [self.layerClient queryControllerWithQuery:query error:&error];
+            if (!self.queryController) {
+                NSLog(@"LayerKit failed to create a query controller with error: %@", error);
```
chipxsd: Same as before.

Code/Models/ATLConversationDataSource.m
```diff
@@ -43,11 +43,14 @@ - (id)initWithLayerClient:(LYRClient *)layerClient query:(LYRQuery *)query
         NSUInteger numberOfMessagesAvailable = [layerClient countForQuery:query error:nil];
         NSUInteger numberOfMessagesToDisplay = MIN(numberOfMessagesAvailable, ATLQueryControllerPaginationWindow);
     
-        _queryController = [layerClient queryControllerWithQuery:query];
+        NSError *error = nil;
+        _queryController = [layerClient queryControllerWithQuery:query error:&error];
+        if (!_queryController) {
+            NSLog(@"LayerKit failed to create a query controller with error: %@", error);
```
chipxsd: same

Code/Controllers/ATLConversationListViewController.m
kmahal: Added a return if nil

Code/Controllers/ATLConversationListViewController.m
kmahal: We don't handle the error below either.  I can return early.

Code/Models/ATLConversationDataSource.m
kmahal: We return an object even if the below execution fails with error, what else should we change here then?

PR #951, opened 2015-09-03
Don't crash if a ATLMessageCollectionViewCell subclass doesn't set this constraint 

sammcd:
I think there are times where a subclass may not define a bubbleWithAvatarLeadConstraint, in those cases I think the application should not crash.

(Reopening this as a new PR because old one was based on my master branch :/)

### DIFF
--- a/Code/Views/ATLMessageCollectionViewCell.m
+++ b/Code/Views/ATLMessageCollectionViewCell.m
@@ -413,11 +413,16 @@ CGFloat const ATLAvatarImageTailPadding = 7.0f;
     if (shouldDisplayAvatarItem) {
         if ([constraints containsObject:self.bubbleWithAvatarLeadConstraint]) return;
         [self.contentView removeConstraint:self.bubbleWithoutAvatarLeadConstraint];
-        [self.contentView addConstraint:self.bubbleWithAvatarLeadConstraint];
+        if (self.bubbleWithAvatarLeadConstraint != nil) {
+            [self.contentView addConstraint:self.bubbleWithAvatarLeadConstraint];
+        }
     } else {
         if ([constraints containsObject:self.bubbleWithoutAvatarLeadConstraint]) return;
         [self.contentView removeConstraint:self.bubbleWithAvatarLeadConstraint];
-        [self.contentView addConstraint:self.bubbleWithoutAvatarLeadConstraint];
+
+        if (self.bubbleWithoutAvatarLeadConstraint != nil) {
+            [self.contentView addConstraint:self.bubbleWithoutAvatarLeadConstraint];
+        }
     }
     [self setNeedsUpdateConstraints];
 }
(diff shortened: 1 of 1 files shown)


sammcd:
Any thoughts on this? Let me know if there is anything I can do to help this get pulled in.

Code/Views/ATLMessageCollectionViewCell.m
```diff
@@ -413,11 +413,16 @@ - (void)shouldDisplayAvatarItem:(BOOL)shouldDisplayAvatarItem
...
         [self.contentView removeConstraint:self.bubbleWithoutAvatarLeadConstraint];
-        [self.contentView addConstraint:self.bubbleWithAvatarLeadConstraint];
+        if (self.bubbleWithAvatarLeadConstraint != nil) {
+            [self.contentView addConstraint:self.bubbleWithAvatarLeadConstraint];
+        }
     } else {
         if ([constraints containsObject:self.bubbleWithoutAvatarLeadConstraint]) return;
         [self.contentView removeConstraint:self.bubbleWithAvatarLeadConstraint];
-        [self.contentView addConstraint:self.bubbleWithoutAvatarLeadConstraint];
+
```
kmahal: extra space here

kmahal:
Hi @sammcd apologies for the late response.  Added one comment.  Additionally, there would be more work to be done to make this work correctly.  These cells are designed to have the constraints specified, and could lead to layout issues without them.  In what case would you not want to set these, as otherwise the avatar image view would have broken constraints.

sammcd:
I have fixed the spacing.

My case is that I have a custom cell subclass that supports a custom mimetype, which atlas handles *very* well. However when I updated to the latest atlas, this new requirement was added that forces all subclasses to set a layout constraint to a _hidden_ property. This seems like a strange practice. My current change simply fixes crashes caused by this new requirement. 

I think there are lot of ways to move forward to remove this requirement, but I also think this change provides value with minimal risk. One thing I could is log some sort of warning in this case alerting subclasses they they should set this constraint. We could also create an alternative api for subclasses to layout this view. 

Let me know.

kmahal:
Hey @sammcd apologies for the late response again.  I just tested this out - if a subclass does not set these constraints, the cells do not render correctly.  We can discuss adding them to the header file, but either way these are important constraints for correct layout.

sammcd:
The use case here is subclassing the cell to make my own custom cell. My custom cell may or may not need to change the leading constraint depending on an avatar. In my current use case my custom cell does not need this constraint to change, therefore there is no issue with layout, however there is a crash without my fix.

Basically — not every custom cell requires these properties, so I believe it makes more sense to avoid crashing to build for the case that the properties are not set. The extent of this change is to prevent crashes — and in the worst case, you will get a layout issue instead of a crash (which is the current status).

kmahal:
HI @sammcd if we get this rebased on master and mergeable we'll get it in.  Let me know if you're still on this, or else we can close it out.  Thanks!

kcoleman731:
@sammcd going to close this PR due to no response. Let me know if you want to rebase and get it merged it.

PR #952, opened 2015-09-03
Fixed broken tests

kmahal:


Tests/ATLConversationListViewControllerTest.m
```diff
@@ -52,15 +52,14 @@ - (void)setUp
 
 - (void)tearDown
 {
+    [super tearDown];
+    [tester waitForAnimationsToFinish];
     [self.testInterface dismissPresentedViewController];
-    self.viewController.queryController = nil;
-    self.viewController = nil;
+    if (self.viewController) self.viewController = nil;
```
chipxsd: Let's try to avoid single line expressions.
```objective-c
if (self.viewController) {
    self.viewController = nil;
}
```

PR #955, opened 2015-09-07
Added a nil safety-check to shouldDisplayAvatarItemAtIndexPath

varkor:
This prevents errors being thrown when avatar items are enabled and a
message sender has a name rather than a userID (such as when a message
is sent by a platform).


kmahal:
Thanks @varkor

PR #960, opened 2015-09-10
Improved long-press behaviour on ATLMessageBubbleView

varkor:
This includes four changes that generally improve the experience of long-pressing on message bubbles.

1. When setting custom menu items on a ATLMessageBubbleView (via `conversationViewController:configureCell:forMessage:`), they are now properly cleared from the global UIMenuController after the long-press. This avoids the following scenario:
Message bubble with extra menu option:
image
Then long-press in the text input field:
image

2. When you long press on a message bubble, and the menu comes up, and then you try panning (without ever lifting your finger from the screen), the scroll position is now locked until you lift your finger from the screen. This replicates the Apple Messenger behaviour.

3. If the message bubble long-press menu items are set to nil or an empty array (again, using `conversationViewController:configureCell:forMessage:`), you are now not prevented from scrolling when the menu would originally have appeared.

4. Added a 100 millisecond appearance/disappearnce transition for the dark alpha mask that covers the message bubble view when long-pressed, to make it more aesthetically-pleasing and replicate the Apple Messenger behaviour.

### DIFF
--- a/Code/Views/ATLMessageBubbleView.m
+++ b/Code/Views/ATLMessageBubbleView.m
@@ -333,7 +333,10 @@ typedef NS_ENUM(NSInteger, ATLBubbleViewContentType) {
             menuController.arrowDirection = UIMenuControllerArrowDefault;
         }
 
+        self.panGestureRecognizer.enabled = NO;
         [menuController setMenuVisible:YES animated:YES];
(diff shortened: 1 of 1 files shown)


kmahal:
@varkor apologies for the late response.  This looks good; it just needs to be rebased on master and then we can merge it.

varkor:
Okay—the branch should be up to date with master now!

PR #969, opened 2015-09-17
Added in uisplitviewcontroller support for the input accessory view's frame

kmahal:


chipxsd:
Add a test where the conversation controller is embedded in another view controller, and shrink the parent's view size. Expect the width of the parent view frame to equal the input bar's.

PR #977, opened 2015-09-22
fixes ios9 window bug

kmahal:
this PR fixes a bug that causes the input accessory background to be black (window color) when presenting the `ATLConversationViewController` on screen.


bradytate:
@kmahal Still seeing this issue. What version of Atlas should include this fix?

PR #979, opened 2015-09-22
dismiss keyboard when left button tapped

### LINKED ISSUE #978: Left Accessory Button Doesn't Resign Keyboard Before Presenting Action Sheet
The view just dims and the keyboard remains, and the action sheet seems to be presented behind the keyboard. 

A simple solution seems to just be to call [messageInputToolbar.textInputView resignFirstResponder] in the ATLMessageInputToolbar Delegate method, messageInputToolbar: didTapLeftAccessoryButton:

neil477:
fixes #978

Issue: tap left accessory button while keyboard is visible, Alertsheet appears behind keyboard.

screenshot png

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -554,6 +554,10 @@ static NSInteger const ATLPhotoActionSheet = 1000;
 
 - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTapLeftAccessoryButton:(UIButton *)leftAccessoryButton
 {
+    if (messageInputToolbar.textInputView.isFirstResponder) {
+        [messageInputToolbar.textInputView resignFirstResponder];
+    }
+    
     UIActionSheet *actionSheet = [[UIActionSheet alloc] initWithTitle:nil
                                                              delegate:self
                                                     cancelButtonTitle:ATLLocalizedString(@"atl.conversation.toolbar.actionsheet.cancel.key", @"Cancel", nil)


neil477:
not sure if you're bumping the min version of atlas to ios 8, if so we stop using alertsheet and just present an alertcontroller

Code/Controllers/ATLConversationViewController.m
```diff
@@ -554,6 +554,7 @@ - (BOOL)shouldDisplayAvatarItemAtIndexPath:(NSIndexPath *)indexPath
 
 - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTapLeftAccessoryButton:(UIButton *)leftAccessoryButton
 {
+    [self.messageInputToolbar.textInputView resignFirstResponder];
```
kmahal: instead of `self` just use the `messageInputToolbar` instance passed in the delegate.  also check `if (messageInputToolbar.isFirstResponder)` before calling it.

kmahal:
One small comment on it, and then we'll get this merged in.  We are still considering how far we want to remain backwards compatible per your `UIAlertController` suggestion.

neil477:
@kmahal oops my bad yeah fixed it based on your suggestions.

kmahal:
Thanks. The check should be if (messageInputToobar.inputTextview.isFirstResponder) since that's what will resign FR status. Apologies for the mix up.

neil477:
@kmahal ahh embarassed i didnt see that, ok checked first responder on the textView.

haemi:
do you know when this will be available via cocoapods?

haemi:
the location button still doesn't work in iOS 9

neil477:
@haemi can you give more context? does it not work when the keyboard is open or just doesnt work in general? I can't reproduce this problem on iOS 9. maybe create a new issue?

haemi:
for me, nothing happens at all, no matter if the keyboard is present or not. Doesn't work in Simulator nor on device.

On an iOS 7 device it works as expected.

neil477:
@haemi hmm for sanity check you could check on your phone you've given your app location permissions?

haemi:
in iOS 9, it's not listed there and I'm not asked (which I am on the iOS 7 device)

chlebta:
@kmahal Even using `UIAlertController` the Action sheet will appear behind keyboard is visible

PR #987, opened 2015-09-30
Adding shouldDisplaySearchController property to ATLConversationListViewController

kmahal:
This PR makes displaying the search bar and search display controller optional by adding the `shouldDisplaySearchController` to the `ATLConversationListViewController`.


cashmash:
the variable is beeing ignored (Atlas 1.0.13)
I set it to NO but it still shows it.

PR #996, opened 2015-10-07
Allow avatars to be shown in the ATLAddressBarViewController

dfinzer:
@kmahal This allows avatars to be shown in the address bar view controller. However, it changes the look of the address bar view controller since I am reusing ATLParticipantTableViewCells, so I wanted to get your advice on whether or not this is the best path to go down.

### DIFF
--- a/Code/Controllers/ATLAddressBarViewController.h
+++ b/Code/Controllers/ATLAddressBarViewController.h
@@ -105,6 +105,15 @@
  */
 @property (nonatomic) ATLAddressBarView *addressBarView;
 
+///----------------------
+// @name UI Configuration
+///----------------------
+
+/**
+ @abstract Whether the view controller should show avatars in the participant list. Default is NO.
+ */
+@property (nonatomic) BOOL shouldShowParticipantAvatars;
+
 ///------------------------------------
 // @name Managing Participant Selection
 ///------------------------------------
--- a/Code/Controllers/ATLAddressBarViewController.m
+++ b/Code/Controllers/ATLAddressBarViewController.m
@@ -22,6 +22,7 @@
 #import "ATLConstants.h"
 #import "ATLAddressBarContainerView.h"
 #import "ATLMessagingUtilities.h"
+#import "ATLParticipantTableViewCell.h"
 
 @interface ATLAddressBarViewController () <UITextViewDelegate, UITableViewDataSource, UITableViewDelegate>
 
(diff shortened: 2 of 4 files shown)


kmahal:
@dfinzer There are color and font changes that occur this way.  Adding configurable properties to the `ATLParticipantTableViewCell` would perhaps work best.  Additionally, instead of implementing a delegate, adding a public property to `ATLAddressBarController` would work best.

Take a pass at a new implementation and we can discuss from there.

dfinzer:
@kmahal Added configurable properties to match existing UI. Let me know how this looks.

dfinzer:
Any updates on whether this looks ok?

dfinzer:
@kmahal @blakewatters any updates on whether this looks okay?

dfinzer:
@kmahal just want to check in on this. Thanks!

dfinzer:
@kmahal any chance we could get this PR in? Been waiting for quite some time.

Code/Controllers/ATLAddressBarViewController.h
```diff
@@ -105,6 +105,12 @@
  */
 @property (nonatomic) ATLAddressBarView *addressBarView;
 
+///----------------------
+// @name UI Configuration
+///----------------------
+
+@property (nonatomic) BOOL shouldShowParticipantAvatars;
```
kmahal: property should be set in cell initializer as default `NO`, as well as header docs explaining what this property does when set.

kmahal:
@dfinzer I've added one comment to fix.  Once complete I'l pull down the branch and verify its behavior again and then we can get it merged.  Thank you for this PR!

Code/Controllers/ATLAddressBarViewController.h
dfinzer: I added a doc, but since this is a view controller property I'm not sure what you mean by setting it in the "cell initializer"

Code/Controllers/ATLAddressBarViewController.h
```diff
@@ -105,6 +105,15 @@
...
 @property (nonatomic) ATLAddressBarView *addressBarView;
 
+///----------------------
+// @name UI Configuration
+///----------------------
+
+/**
+ @abstract Whether the view controller should show avatars in the participant list.
+ */
+@property (nonatomic) BOOL shouldShowParticipantAvatars;
```
kmahal: The default value of this property is not set in `viewDidLoad` currently.  The default value should be set to `NO` and documented.

dfinzer: Ah, got it. Set the default value and added documentation

PR #997, opened 2015-10-07
Avatar with Participants Initials

alexmaie:
The participants initial are showed until the avatar id downloaded from the server

### DIFF
--- a/Code/Views/ATLAvatarImageView.m
+++ b/Code/Views/ATLAvatarImageView.m
@@ -110,7 +110,13 @@ NSString *const ATLAvatarImageViewAccessibilityLabel = @"ATLAvatarImageViewAcces
 - (void)setAvatarItem:(id<ATLAvatarItem>)avatarItem
 {
     if ([avatarItem avatarImageURL]) {
-        self.initialsLabel.text = nil;
+        //fallback if avatar is not available on the server
+        if(avatarItem.avatarInitials){
+            self.initialsLabel.text = avatarItem.avatarInitials;
+        }else{
+            self.initialsLabel.text = nil;
+        }
+        //load avatar from server
         [self loadAvatarImageWithURL:[avatarItem avatarImageURL]];
     } else if (avatarItem.avatarImage) {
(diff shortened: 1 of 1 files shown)


Code/Views/ATLAvatarImageView.m
```diff
@@ -110,7 +110,13 @@ - (void)dealloc
 - (void)setAvatarItem:(id<ATLAvatarItem>)avatarItem
 {
     if ([avatarItem avatarImageURL]) {
-        self.initialsLabel.text = nil;
+        //fallback if avatar is not available on the server
```
kmahal: Why would this be a fallback if the avatar is not available on the server?  It will be making a network call to the server, and the initials are only temporary.

Code/Views/ATLAvatarImageView.m
```diff
@@ -110,7 +110,13 @@ - (void)dealloc
...
+        if(avatarItem.avatarInitials){
+            self.initialsLabel.text = avatarItem.avatarInitials;
```
kmahal: After the image is downloaded, the initials are still shown.  We wouldn't want that behavior.

kmahal:
@alexmaie I've added a few comments.  We're also discussing internally if we would want this behavior or not, as not all users would want avatar initials to show up during load.  Finding a way to make this behavior optional would be optimal.

alexmaie:
Hey, 

So basically I will tell you the use case for our current project.
The avatar is an optional element for a participant in our case. But this we can find out after we try to download the avatar and the url request with a 404 http code.

kmahal:
@alexmaie This behavior is not one we would want to accommodate for in Atlas, as it is not the likely use case for most of our users.  The best approach may be forking Atlas for your own needs, or as suggested finding an implementation that allows for customization.

alexmaie:
Ok,

Thanks for the heads up

kmahal:
@alexmaie if you are going to keep it forked instead of a PR, feel free to close this out then.

PR #1003, opened 2015-10-10
load image from url after setting placeholder image or avatar initials

neil477:
@kmahal another take on #997 . We needed similar behaviour in our app to this user, thought this might be more generally applicable. Basically we set the image or avatar initials if available and use that as a place holder and go get the image from the server. I also removed a parameter that looked redundant to me in the update image method.

### DIFF
--- a/Code/Views/ATLAvatarImageView.m
+++ b/Code/Views/ATLAvatarImageView.m
@@ -109,15 +109,16 @@ NSString *const ATLAvatarImageViewAccessibilityLabel = @"ATLAvatarImageViewAcces
 
 - (void)setAvatarItem:(id<ATLAvatarItem>)avatarItem
 {
-    if ([avatarItem avatarImageURL]) {
-        self.initialsLabel.text = nil;
-        [self loadAvatarImageWithURL:[avatarItem avatarImageURL]];
-    } else if (avatarItem.avatarImage) {
-        self.initialsLabel.text = nil;
-        self.image = avatarItem.avatarImage;
+    //set image or initials if available
+    if (avatarItem.avatarImage) {
+        [self populateImageWithImage:avatarItem.avatarImage];
     } else if (avatarItem.avatarInitials) {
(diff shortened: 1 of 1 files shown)


Code/Views/ATLAvatarImageView.m
```diff
@@ -109,19 +109,38 @@ - (void)dealloc
 
 - (void)setAvatarItem:(id<ATLAvatarItem>)avatarItem
 {
+    //set image or initials initialls
+    [self populateAvatarImageOrInitialsForAvatarItem:avatarItem];
```
kmahal: instead of breaking it out into this method and several others, you can keep the implementation in this method, with the new default behavior you'd like to see.

neil477: I just thought it might be easier to read if I broke it out into multiple functions. I'll update it to just be in one function though!

kmahal:
@neil477 any update on this?

neil477:
@kmahal will fix tonight sorry

neil477:
@kmahal fixed sorry for the delay

Code/Views/ATLAvatarImageView.m
```diff
@@ -109,19 +109,30 @@ - (void)dealloc
...
-        self.initialsLabel.text = nil;
-        self.image = avatarItem.avatarImage;
-    } else if (avatarItem.avatarInitials) {
-        self.image = nil;
-        self.initialsLabel.text = avatarItem.avatarInitials;
     }
     _avatarItem = avatarItem;
 }
 
+-(void)populateImageWithImage:(UIImage *)image {
```
kmahal: group these 2 methods above the `loadAvatarImageWitihURL` method.  Add a pragma mark as well to describe this group of methods.

kmahal:
@neil477 one small comment and then we can merge.

neil477:
@kmahal added a comment too just to document the assumption

PR #1004, opened 2015-10-10
add support for displaying system user name and avatar

neil477:
@kmahal we make heavy use of sending automated system messages to our users. The system user's name comes as unknown user and avatar is blank. This should allow users of Atlas to set system user avatar and name. Feedback appreciated, would be happy to modify if it needed.


kmahal:
@neil477 thanks for the PR.  System messages aren't participants in a conversation, thus we are hesitant to have a return type adhere to `ATLParticipant`.  Ideally we would expose the `LYRActor` object on a conversation's participants instead of `participantID` strings which would let us use one delegate for both cases.  Let me re-think an implementation here and get back to you.

kmahal:
@neil477 I'm going to close this out for now - there are forthcoming changes to `LayerKit` in the coming weeks that will address this and impact Atlas.  Instead of introducing this and changing it soon, feel free to fork Atlas if you require a fix immediately, and I'll update this ticket when we begin the future work.

PR #1022, opened 2015-10-15
Added ATLMessageInputToolbar right button title property

kmahal:
Makes the right accessory button title configurable, and resizes the textinput view according to the new text.


chipxsd:
Nice work, looks good! One last thing though, can you add a short test that verifies the width of the button frame and the frame of the input bar when the text changes?

garhiggins:
Awesome!

PR #1029, opened 2015-10-16
Make ATLConversationView a public property

dfinzer:
I need this to be public so I can override MessageInputToolbar. Does this change seem reasonable?


kmahal:
@dfinzer there is another open PR: https://github.com/layerhq/Atlas-iOS/pull/946 that will be merged soon with this functionality that allows for implicit use of the Atlas implementation without needing to set the input accessory view yourself.  I'm going to close this PR for now, and you should see an update on that thread soon.

PR #1033, opened 2015-10-19
Remove hard-code circle radius treatment for AvatarItem 

charlespsdowd:
Added optional property to ATLAvatarItem protocol (`avatarRadius:CGFloat`)
Added radiusOverride property to ATLAvatarImageView (`radiusOverride:CGFloat`)
When you wnat to overide the radius on the avatar you set this value to > 0, otherwise you get the default circle
The closest to a square you can get is a cornerRadius of 1


Code/Protocols/ATLAvatarItem.h
```diff
@@ -42,4 +42,11 @@
  */
 @property (nonatomic, readonly) NSString *avatarInitials;
 
+@optional
+/**
+ @abstract Returns the avatar image radius for display. 0 makes this the default circle.
+ */
+@property (nonatomic, readonly) CGFloat avatarRadius;
```
kmahal: This is a property for `ATLAvatarImageView` not for `ATLAvatarItem`.

Code/Views/ATLAvatarImageView.m
```diff
@@ -24,6 +24,7 @@ @interface ATLAvatarImageView ()
 
 @property (nonatomic) UILabel *initialsLabel;
 @property (nonatomic) NSURLSessionDownloadTask *downloadTask;
+@property (nonatomic) CGFloat radiusOverride;
```
kmahal: This is not a clean implementation.  The view already has:
```objc
/**
 @abstract Sets the diameter for the avatar image view. Default is 30.
 @discussion Bounds for the image view are clipped to half of the diameter to create a circular
 image.
 */
@property (nonatomic) CGFloat avatarImageViewDiameter UI_APPEARANCE_SELECTOR;
```

What use case are you trying to achieve?  a square?  If necessary to change the circle/the corner radius then add a property here that is set internally and configured by changing `avatarImageViewDiameter`, but can be overridden if necessary.

charlespsdowd: I want to change the avatar shape depending on the type of Avatar. The current implementation forces everything to be a circle or the same...very specific to my needs but it's the lack of access to the circle avatar I wanted to find a fix for. Happy to go in another direction for this!

kmahal:
@charlespsdowd look at the delegate method `conversationViewController:configureCell:forMessage:` that will let you configure each cell on a per-cell basis.  This way you can get the radius customized as you see fit.  LMK if that works and we can close this out.

charlespsdowd:
I love what yo've done with Atlas+Layer. I really want to be able to contrbute and continue to upgrade the code base, so I'm motivated to get the right solution.
I will look at that and make that work. I was hoping to not have to rebuild all of the message rendering just to overcome the current hard-coded avatar cropped to a circle.

kmahal:
@charlespsdowd thanks for the kind words.  You wouldn't have to rebuild it all, just implement your custom logic here for how you want to size it. UI_Appearance is great for a one size fits all approach, but your needs will require the use of additional customization that is allowed for within our APIs.  We can re-open this ticket if the solution doesn't work.

charlespsdowd:
I still need to do this for every View subclass that uses the AvatarItem
1. ConversationListViewController
2. ConversationViewController
3. ParticipantViewController
I know I must be missing something obvious here but the radius (not the diameter) is hard-coded in the AvatarImageView???

PR #1036, opened 2015-10-21
APPS-2109 - Slow Scrolling Performance

chipxsd:
The issue was the way how auto layout engine tries to calculate the constraints using `NSLayoutRelationLessThanOrEqual` by doing multiple passes of sizing in order to satisfy a constraint (it starts with a 2-pass approach, and if the value still changes the second time, it does another 2-pass attempt). And while it does this, it calculates the constraints of the whole layout graph (since our layout hierarchy does not break at any point, it goes from the cell's label all the way up to the UIWindow).
I eliminated the constraint that would cause this guess the number approach and manually size the constraint based on the calculated text size.


JRam13:
I'm still experiencing some bad scrolling performance issues, not to mention a whole bunch of NSLayoutConstraint warnings.

PR #1042, opened 2015-10-28
Adding delete message option

chlebta:
Adding delete message option on bubble long click. this work only for
`LYRMessage`.
TODO fix delete function in `LYRMockContentStore` to avoid crash app


kmahal:
@chlebta you can customize `UIMenuController` actions by implementing the delegate `conversationViewController:configureCell:forMessage:` and configuring the cell options with the actions you'd like to see.

PR #1050, opened 2015-11-10
ATLMessageInputToolbar call convertRect:toView: from superview

jeffdgr8:
This error is not immediately obvious since the superview's frame.origin.x is the same as the containerViewController view's frame.origin.x. But this calculation should be done relative to the superview, since this is the coordinate system the containerViewController's frame is relative to.

### DIFF
--- a/Code/Views/ATLMessageInputToolbar.m
+++ b/Code/Views/ATLMessageInputToolbar.m
@@ -132,7 +132,7 @@ static CGFloat const ATLButtonHeight = 28.0f;
     
     // This makes the input accessory view work with UISplitViewController to manage the frame width.
     if (self.containerViewController) {
-        CGRect windowRect = [self.containerViewController.view convertRect:self.containerViewController.view.frame toView:nil];
+        CGRect windowRect = [self.containerViewController.view.superview convertRect:self.containerViewController.view.frame toView:nil];
         frame.size.width = windowRect.size.width;
         frame.origin.x = windowRect.origin.x;
     }


kmahal:
@jeffdgr8 can you explain further?  Is this causing a specific bug for you?  The reasoning for using the containerViewController's view was to define the width of the input bar no matter where it's placed.  The Atlas Messenger project sets up the project in a `UISplitViewController`, which allows for the input accessory view to just remain the width of the containerViewController, like how iMessage functions across the 6s+ and iPad.  If the `ATLConversationViewController` is added as a child view controller and arbitrarily placed at an origin not aligned with the superview, then this could get thrown off.

Let me know if there is something I'm missing here.

jeffdgr8:
Yes, I understand what it's doing. I came across the bug because I'm doing something similar to also allow a subclassed `ATLMessageInputToolbar` to appear above the tab bar which contains our `UISplitViewController`. But the way you are calculating the `convertRect:toView:` relative to the toolbar's own view is incorrect. The toolbar view's frame is relative to it's superview, so this method should be called from the superview. It's not obvious that it's incorrect because both calculations come out to the same result in this case, specifically because the `ATLConversationViewController`'s frame.origin.x is 0.

In my case, how I discovered the bug, when the nav bar is set to non-translucent it causes the `ATLConversationViewController`'s frame.origin.y to be 64, which is offset from its superview's frame.origin.y. So calling this method from the `ATLConversationViewController`'s view to get the rect in screen coordinates ends up doubling the origin.y value to 128 because its being called incorrectly from the frame's own view. Calling it correctly from the superview returns the correct screen coordinate results.

This would only be obvious as a bug in your use case if the `ATLConversationViewController` had an x offset set on its frame so that it wasn't equal to its superview's x offset. So even though its not causing an issue as is with the way its frame is set in the `UISplitViewController`, it should still be corrected.

kmahal:
@jeffdgr8 apologies for the late response.  I'll test this behavior out in the AM and merge it in.  One thing to note - this container view's frame is only used to calculate the x of the input toolbar, not the y.  How would this code calculate the incorrect y coordinate as well?

jeffdgr8:
It doesn't affect the y coordinate in your use. I just used the same line of code in order to calculate the y coordinate and since there is an offset in the y coordinate when setting the nav bar to translucent, this is how I discovered that the calculation is incorrect. Again, you'll only see an issue if the frame has a non-zero origin. In your use, a non-zero origin.x would be an issue. In my use, a non-zero origin.y caused the issue.

kmahal:
Thanks @jeffdgr8!

PR #1076, opened 2016-01-04
Feature/ATLMessageToolbar custom right button

ricardorover:
This will allow:
- change right button appearance and behavior when in send mode (when displaysRightAccessoryImage is false or textInputView is not empty). Default behavior is set text to atl.messagetoolbar.send.key.
- change right button appearance and behavior when button is in 'custom action' mode (when displaysRightAccessoryImage is true and textInputView is empty). Default behavior is set button image to rightAccessoryImage.

This pull request comes from #946, and in my project this is a required feature because we need to set Send button as a image (not a text), and other than have a location button we have a voice message button. The original code in ATLMessageToolbar foresees a text button as Send and a image button as Location (or any other custom action), thus not allowing a image button in both cases.

In first pull request I made public the configureRightAccessoryButtonState method, but after putting some thoughts in that issue, I realised that allowing configuration for each state (send and custom action buttons) looks more integrated with original code (that has displaysRightAccessoryImage flag to enable or disable this custom state).

I made a separated pull request that we can discuss better before merging.


kmahal:
Thanks for the PR @ricardorover.  Instead of opening up this customization as 2 separate methods, for state of 0 input and with 1+ items, consider adding an optional delegate method to the `ATLMessageInputToolbarDelegate` and moving the protocol from the `ATLConversationViewController.m` to the `.h` file.  A delegate such as `- (BOOL)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar configureStateForTextInputView:(ATLMessageComposeTextView *)textInputView` would pass all relevant information to the delegate and let them choose to customize the input bar (button included) as they see fit, instead of forcing them into those 2 methods.

ricardorover:
Nice suggestion, @kmahal. I changed the code to use the delegate way, see if the code and the docs I added are ok.

Code/Views/ATLMessageInputToolbar.m
```diff
@@ -371,16 +371,18 @@ - (void)acceptAutoCorrectionSuggestion
 
 - (void)configureRightAccessoryButtonState
 {
-    if (self.textInputView.text.length) {
-        [self configureRightAccessoryButtonForText];
-        self.rightAccessoryButton.enabled = YES;
-    } else {
-        if (self.displaysRightAccessoryImage) {
-            [self configureRightAccessoryButtonForImage];
+    if (![self.inputToolBarDelegate respondsToSelector:@selector(messageInputToolbar:configureStateForTextInputView:)] || [self.inputToolBarDelegate messageInputToolbar:self configureStateForTextInputView:self.textInputView]) {
```
kmahal: This wouldn't be correct design, because you first want to check to see if it responds to the delegate before calling it, thus the `||` is misplaced.  Instead, do something like this:
```objc
 - (void)configureRightAccessoryButtonState
{
  if ([self.inputToolBarDelegate respondsToSelector:@selector(messageInputToolbar:configureStateForTextInputView:)] {
    BOOL shouldDefaultConfiguration = [self.inputToolbarDelegate messageInputToolbar:self configureStateForTextInputView:self.textInputView];
    if (shouldDefaultConfiguration) {
      [self defaultRightButtonConfiguration];
    }
  }
} else {
   [self defaultRightButtonConfiguration];
}

- (void)defaultRightButtonConfiguration
{
// default config here
}
```

Looking at the design, the 2nd part of the delegate isn't required since the input toolbar already has the right button and text input view as public properties, so just passing down the input toolbar would work.

Code/Views/ATLMessageInputToolbar.h
```diff
@@ -50,6 +50,13 @@ extern NSString *const ATLMessageInputToolbarAccessibilityLabel;
 @optional
 
 /**
+ @abstract Implement this method to customize toolbar behavior and appearance. It is called on setup, when textInputView value changes and when media is attached.
+ @discussion If not implemented or NO is returned, default behavior is to configure right button with atl.messagetoolbar.send.key label if textInputView has text value or image attached, or show button with rightAccessoryImage otherwise.
```
kmahal: discussion and result differ.  `YES` should do default config, `NO` should let the user override it. Additionally there is no need to point out what key is used since it's in the other part of the docs.

Code/Views/ATLMessageInputToolbar.m
ricardorover: I will do the changes you suggest, but it's only code styling, right? The delegate method is only called if respondsToSelector is true.
I couldn't find a similar snippet in Atlas code, so I implemented that way, but I'm happy to adjust the code style.

I agree about the 2nd part of the delegate. I was following your suggestion, I have not spent much time thinking about it. I will adjust that as well.

Code/Views/ATLMessageInputToolbar.h
ricardorover: Sorry, I forgot to fix the discussion text.

I couldn't find another reference to "atl.messagetoolbar.send.key" in docs, only in the code. Should I just remove the key from the discussion text, or do I remove the discussion itself? Since I wrote the discussion to describe the default behavior

Code/Views/ATLMessageInputToolbar.m
kmahal: I believe so, but it's always good to check if something responds to a selector before calling it anyways, since at that point you could just not use the check.

Code/Views/ATLMessageInputToolbar.h
kmahal: just remove the key and we'll be good!

kmahal:
@ricardorover have been doing some Atlas cleanup today and dropped a few comments!

ricardorover:
@kmahal I pushed the changes, let me know if there is yet any adjust to be made.

PR #1089, opened 2016-01-22
Feature/sync delete

kcoleman731:
* Adds support for synchronized delete.


Code/Controllers/ATLConversationListViewController.m
```diff
@@ -49,7 +49,7 @@ @implementation ATLConversationListViewController
 NSString *const ATLConversationListViewControllerTitle = @"Messages";
 NSString *const ATLConversationTableViewAccessibilityLabel = @"Conversation Table View";
 NSString *const ATLConversationTableViewAccessibilityIdentifier = @"Conversation Table View Identifier";
-NSString *const ATLConversationListViewControllerDeletionModeLocal = @"Local";
+NSString *const ATLConversationListViewControllerDeletionModeLocal = @"Sync";
```
blakewatters: Change variable name from containing `Local`. Instead of `Sync` I'd go with `My Devices`. Maybe we should change `Global` to `Everyone` ?

kcoleman731: works

PR #1093, opened 2016-02-03
Fixing query controller execution issue

kmahal:
This PR fixes an issue with the tableview dequeueing a cell that hasn't been registered.

### DIFF
--- a/Code/Controllers/ATLConversationListViewController.m
+++ b/Code/Controllers/ATLConversationListViewController.m
@@ -86,6 +86,7 @@ NSString *const ATLConversationListViewControllerDeletionModeEveryone = @"Everyo
     _allowsEditing = YES;
     _rowHeight = 76.0f;
     _shouldDisplaySearchController = YES;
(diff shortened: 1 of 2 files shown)


Code/Controllers/ATLConversationListViewController.m
```diff
@@ -311,6 +312,9 @@ - (void)reloadCellForConversation:(LYRConversation *)conversation
     if (!conversation) {
         @throw [NSException exceptionWithName:NSInvalidArgumentException reason:@"`conversation` cannot be nil." userInfo:nil];
     }
+    if (!self.queryController) {
```
kcoleman731: Are we sure we want to just return nil here? Why not assert in the case there is no query controller?

kmahal: Yea I think so.  Query controller is a property defined outside of the context of the method, and I think this should be a soft failure instead of a hard exception, since calling it early with a valid conversation shouldn't be exceptional.

PR #1103, opened 2016-02-16
Allow client to specify that names be shown even in two person conversations

klep:


kmahal:
Already available via:
```objc
/**
 @abstract A Boolean value that determines whether or not an avatar is shown if there is only one other participant in the conversation.
 @default `NO`.
 Should be set before `[super viewDidLoad]` is called.
 */
@property (nonatomic) BOOL shouldDisplayAvatarItemForOneOtherParticipant;

/**
 @abstract A Boolean value that determines whether or not an avatar is shown next to the outgoing messages
 @default `NO`.
 */
@property (nonatomic) BOOL shouldDisplayAvatarItemForAuthenticatedUser;
```

klep:
Unless I'm missing something, that flag will not display the sender label due to this check:

```
if (self.conversation.participants.count <= 2) return NO;
```

kmahal:
Apologies, this is about the naming not avatars.  I'll reopen and check it soon.

kmahal:
@klep this looks good.  Let's get a test in `ATLConversationViewControllerTest` for this behavior and then we'll be good to merge.

klep:
@kmahal: Test added!

kmahal:
simulator screen shot feb 23 2016 5 36 09 pm

@klep the test looks good, however if the controller does not also set `shouldDisplayAvatarItemForOneOtherParticipant` true then the above occurs.  The name isn't aligned with the bubble in this case.  Let me know when you fix that functionality and we can discuss.

klep:
Good catch. Do you think it's better to make it one flag? I'm not sure why you'd have one without the other, and we don't have that level of granularity on showing avatars or labels when there are > 2 participants. I'd suggest either always showing the label if we're showing the avatar, or renaming the flag shouldDisplaySenderInfoForOneOtherParticipant.

kcoleman731:
@klep picking this up for Kabir. Did you guys sort this out or still pending decision?

klep:
I don't think we ever did. I'm not sure I understand why this distinction exists between conversations with 1 person or multiple people. Seems like in all cases, you'd want the name and avatar, hence the proposed flag. But I'll defer to you guys -- I don't have a problem maintaining our own version of Atlas to get this functionality.

Issue #1083, opened 2016-01-14
editable ATLAddressBarViewController and how to self resize?


jessearmand:
After using it for a while in a group conversation, I've noticed that:

1. Once a conversation is created, the address bar is disabled by default, even though it's possible to add new participant to a conversation?

2. The address bar refused to display all participants name if they don't fit a specific size **when** it's in a disabled state. But, it can display all participants and self-resize, if I manually set `selectedParticipants` after a conversation is created. 

    However, once I do this, the address bar automatically becomes the default first responder, and I can't find any way to resign the address bar text view from first responder status. I suppose the address bar is not editable by default once a conversation is created? Which leads to my next question.

3. There are few situations where the address bar becomes disabled, is it possible to override this, and make it remain editable? What was the reason for disabling it?

4. The `disabledStringForParticipants:` will fallback to displaying the number of participants, if their names don't fit the size of the address bar. But, this is misleading to users, as it display participants count **minus** the first participant.

kmahal:
@jessearmand :
1. This is just like the iMessage experience.  Look at Atlas Messenger (https://github.com/layerhq/Atlas-Messenger-iOS) for how we add or remove participants.
2. The address bar is not editable directly after a conversation is created, just like iMessage.
3. See point 1, it is after the iMessage aesthetic.
4. What do you mean first participant?

jessearmand:
Take a look at this from `ATLAddressBarViewController.m`:

```objc
- (NSString *)disabledStringForParticipants:(NSOrderedSet *)participants
{
    [self.addressBarView.addressBarTextView layoutIfNeeded]; // Layout text view so we can have an accurate width.
    
    // First participant is removed
    __block NSString *disabledString = [participants.firstObject firstName];
    NSMutableOrderedSet *mutableParticipants = [participants mutableCopy];
    [mutableParticipants removeObject:participants.firstObject];
    
    __block NSUInteger remainingParticipants = mutableParticipants.count;
    [mutableParticipants enumerateObjectsUsingBlock:^(id<ATLParticipant> participant, NSUInteger idx, BOOL *stop) {
        NSString *othersString = [self otherStringWithRemainingParticipants:remainingParticipants];
        NSString *truncatedString = [NSString stringWithFormat:@"%@ %@", disabledString, othersString];
        // Could we just display all of them and resize, just when we set the participants in editable mode?
        if ([self textViewHasSpaceForParticipantString:truncatedString]) {
            remainingParticipants -= 1;
[...]

PR #1112, opened 2016-02-23
Fitting address bar participants data

jessearmand:
Related to #1083 

I haven't concluded on my solution. But, this change just ignores the `textViewHasSpaceForParticipantString`, because it's often inaccurate (the value is often too strict).

So there are two issues that I am trying to solve: 
* remaining participants count
* how to display the most number of participant names


kmahal:
@jessearmand This would lead to a cut off participant list which we do not want.  If there are sizing fixes in `textViewHasSpaceForParticipantString` then by all means submit a PR to fix that, but as it stands this PR would break intended functionality.

jessearmand:
@kmahal Well that is the issue, `textViewHasSpaceForParticipantString` is using `sizeWithAttributes:` which from my experience is rarely reliable to calculate width of multiline-text. 

Nevermind I'm closing this, originally this is an issue that I posted in #1083, just trying to explain more. 

For now, I don't see how the original disabledString is better than cutting it off when it doesn't fit, as they serve the same purpose (just with different labels), and I can workaround it by showing all participants in other areas of the UI.

PR #1114, opened 2016-02-25
Fix bug that could occur with unexpected conversation list tableview

marioradonic:
Bug repro steps:

1. Open conversation list
2. Start searching and filtering participants, make sure there are less filtered participants then initially
3. Go into details of filtered conversation
4. Go back and app crashes with:

```
Fatal Exception: NSRangeException
*** -[__NSOrderedSetM objectAtIndex:]: index 7 beyond bounds for empty ordered set
-[ATLConversationListViewController configureCell:atIndexPath:] (ATLConversationListViewController.m:279)

```
This bug happens because **UITableViewController**, which **ATLConversationListViewController** is a subclass of, reloads data on main tableView, but in the configureCell method there is no tableView checkup, it uses queryController which depends on wether searchController isActive or not.

So if you have less filtered results then unfiltered results it is possible for main tableView to ask for a cell to ask for a cell at row 7, while searchQueryController has only 4 results for example, which causes this crash.
Even if you do have enough filtered results for this method to not crash, there could be some unexpected behavior, like configuring cells of main tableView which is below filtered results tableView.

### DIFF
--- a/Code/Controllers/ATLConversationListViewController.m
+++ b/Code/Controllers/ATLConversationListViewController.m
@@ -108,7 +108,7 @@ NSString *const ATLConversationListViewControllerDeletionModeEveryone = @"Everyo
 - (void)viewDidLoad
 {
     [super viewDidLoad];
-    
+
     self.title = ATLLocalizedString(@"atl.conversationlist.title.key", ATLConversationListViewControllerTitle, nil);
     self.accessibilityLabel = ATLConversationListViewControllerTitle;
 
@@ -116,7 +116,7 @@ NSString *const ATLConversationListViewControllerDeletionModeEveryone = @"Everyo
     self.tableView.accessibilityIdentifier = ATLConversationTableViewAccessibilityIdentifier;
     self.tableView.isAccessibilityElement = YES;
     self.tableView.tableFooterView = [[UIView alloc] initWithFrame:CGRectZero];
-    
+
     if (self.shouldDisplaySearchController) {
         self.searchBar = [[UISearchBar alloc] initWithFrame:CGRectZero];
         [self.searchBar sizeToFit];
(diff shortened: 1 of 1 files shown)


Code/Controllers/ATLConversationListViewController.m
```diff
@@ -255,15 +255,16 @@ - (void)setupConversationDataSource
...
 }
 
 - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
 {
     NSString *reuseIdentifier = [self reuseIdentifierForConversation:nil atIndexPath:indexPath];
-    
+
     UITableViewCell<ATLConversationPresenting> *conversationCell = [tableView dequeueReusableCellWithIdentifier:reuseIdentifier forIndexPath:indexPath];
-    [self configureCell:conversationCell atIndexPath:indexPath];
+
```
kmahal: remove extra line

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -511,6 +513,16 @@ - (LYRQueryController *)queryController
     }
 }
 
+- (LYRQueryController *)queryControllerForTableView:(UITableView *)tableView {
+    if (tableView == self.tableView) {
```
kmahal: this block can just be
```objc
return (tableView == self.tableView) ? _queryController : _searchDisplayController;
```
since it's handled internally and we know what tableviews exist

marioradonic: Little defensive programming never hurt nobody. At this point I can't see why, but someone could be adding the third tableView to this view controller and this is just a little helper to whoever does that.

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -511,6 +512,16 @@ - (LYRQueryController *)queryController
...
+        return _queryController;
+    } else if (tableView == self.searchController.searchResultsTableView) {
+        return _searchQueryController;
+    } else {
+        @throw [NSException exceptionWithName:NSInternalInconsistencyException reason:@"TableView doesn't have query controller" userInfo:nil];
```
kmahal: Instead of `TableView doesn't have query controller` let's go with `Controller does not support additional tableviews." since it's not the query controller, but the table view that would be additional.

kcoleman731:
@marioradonic Picking this one up for Kabir. If you wouldn't mind rebasing, we can look at getting this merged. Thanks!

PR #1123, opened 2016-03-07
Fix swift support

dulaccc:
This optional types are required for Swift compatibility.
There might be more out there to fix, I'm finding them one by one.

Cheers


dulaccc:
I fixed the tests also because I needed them to have a simple way of showing the changes I made for swift compatibility were still compiling correctly

:beers:

kcoleman731:
@dulaccc Apologies for the delay. 

Thank you very much for the PR! Merging away.

PR #1124, opened 2016-03-07
Fix the Carthage build

dulaccc:
As pointed out by the issue [Carthage#1096](https://github.com/Carthage/Carthage/issues/1096) and [Carthage#1077](https://github.com/Carthage/Carthage/issues/1077) : 
The `CFBundlePackageType` value of the Info.plist was `BNDL`, but it needs to be `FMWK`.

Except if `BNDL` has been set on purpose ^^ which I'm about to find out I guess.

Plus it was missing a `CFBundleExecutable` entry in the Info.plist

Cheers

### DIFF
--- a/Resources/Info.plist
+++ b/Resources/Info.plist
@@ -4,6 +4,8 @@
 <dict>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>$(EXECUTABLE_NAME)</string>
 	<key>CFBundleIdentifier</key>
 	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
 	<key>CFBundleInfoDictionaryVersion</key>
--- a/Resources/Info.plist
+++ b/Resources/Info.plist
@@ -11,7 +11,7 @@
 	<key>CFBundleName</key>
 	<string>$(PRODUCT_NAME)</string>
 	<key>CFBundlePackageType</key>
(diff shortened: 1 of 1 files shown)


blakewatters:
Pushed `FMWK` change to master, but frameworks don't have an executable

dulaccc:
Thanks for bringing the fix on the master branch.

True, frameworks don't have executables in the strict sense of it, but when you configure a new **Target** in XCode and choose the Cocoa Framework template, the resulting `Info.plist` contains an `CFBundleExecutable` entry with the default value of `$(EXECUTABLE_NAME)`.

I know Carthage is causing issues due to the high maintenance it requires right now.
Though, Carthage is using this default plist entry to [find the binary](https://github.com/Carthage/Carthage/blob/a165f645bcad6481dde5ddb9a6c26efaf27d4378/Source/CarthageKit/Xcode.swift#L1510-L1512).

Don't you think it worth considering including this entry ?

feifanzhou:
@dulaccc Thanks for giving this a go — if adding `CFBundleExecutable` causes the Carthage build to work, can you fork and send us a sample working Xcode project?

dulaccc:
Of course I can do that,
but don't you think it's obvious from [these lines](https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Xcode.swift#L1510-L1512) ? The binary won't ever be found if there is no such `CFBundleExecutable` entry

I'm sending you the Xcode project in addition :)

PR #1128, opened 2016-03-11
Fixed image message preview for PNGs

sebyddd:
Message cell doesn't display the preview of the image if it's not a JPG. Added PNG fallback in case image is not JPG.


Code/Views/ATLMessageCollectionViewCell.m
```diff
@@ -128,8 +128,14 @@ - (void)presentMessage:(LYRMessage *)message
 
 - (void)configureBubbleViewForTextContent
 {
-    LYRMessagePart *messagePart = self.message.parts.firstObject;
-    NSString *text = [[NSString alloc] initWithData:messagePart.data encoding:NSUTF8StringEncoding];
+   NSMutableString *text = [[NSMutableString alloc] initWithString:@""];
```
kmahal: @sebyddd we guarantee that text is encoded into only 1 part.  let's get a fix for that and we can get it in.

Code/Views/ATLMessageCollectionViewCell.m
```diff
@@ -476,4 +495,5 @@ + (CGSize)cellSizeForImageMessage:(LYRMessage *)message
     return size;
 }
 
+
```
kmahal: remove extra line

Code/Views/ATLMessageCollectionViewCell.m
```diff
@@ -426,14 +432,23 @@ + (CGSize)cellSizeForMessage:(LYRMessage *)message inView:(UIView *)view
 
 + (CGSize)cellSizeForTextMessage:(LYRMessage *)message inView:(id)view
 {
+    if ([[self sharedHeightCache] objectForKey:message.identifier]) {
```
kmahal: not necessary.  we already have this in 
```
+ (CGSize)cellSizeForMessage:(LYRMessage *)message inView:(UIView *)view
{
    if ([[self sharedHeightCache] objectForKey:message.identifier]) {
        return [[[self sharedHeightCache] objectForKey:message.identifier] CGSizeValue];
    }
```

Code/Views/ATLMessageCollectionViewCell.m
sebyddd: @kmahal – The issue is that there are cases where the message is split into different parts, with different encodings. This is reproducible with emojis and non-latin characters. Therefore we need to check for encoding on each part and return the concatenated result.

kcoleman731:
@sebyddd Any chance you can update per Kabirs comments?

PR #1139, opened 2016-03-27
Fix Carthage support.

tomco:
Carthage support seems to be broken at the moment. Fixing it was quite simple: the framework scheme was not marked as shared.

PR #1140, opened 2016-03-27
Incorrect deletionModes type

tomco:
After fixing #1139 I found that using Xcode 7.3 Altlas still won't build using carthage:
```objc
ATLConversationListViewController.m:330:114: error: incompatible pointer types sending 'NSArray<NSNumber *> *' to parameter of type 'NSArray<UITableViewRowAction *> * _Nonnull' [-Werror,-Wincompatible-pointer-types]
        NSArray *customActions = [self.dataSource conversationListViewController:self rowActionsForDeletionModes:self.deletionModes];
                                                                                                                 ^~~~~~~~~~~~~~~~~~

In file included from […]/Atlas-iOS/Code/Controllers/ATLConversationListViewController.m:22:
[…]/Atlas-iOS/Code/Controllers/ATLConversationListViewController.h:130:176: note: passing argument to parameter 'deletionModes' here
- (NSArray *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController rowActionsForDeletionModes:(NSArray < UITableViewRowAction*> *)deletionModes;
```

Seems that the SDK/compiler got more picky or the method signature changed. 

Not sure why using carthage this is an error while compiling using Xcode just gives a warning. I silenced the warning/error for now casting to the correct type for now. Not sure if this is just hiding a bug…

### DIFF
--- a/Code/Controllers/ATLConversationListViewController.m
+++ b/Code/Controllers/ATLConversationListViewController.m
@@ -327,7 +327,7 @@ NSString *const ATLConversationListViewControllerDeletionModeEveryone = @"Everyo
 {
     NSMutableArray *actions = [NSMutableArray new];
     if ([self.dataSource respondsToSelector:@selector(conversationListViewController:rowActionsForDeletionModes:)]) {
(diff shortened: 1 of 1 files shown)


kcoleman731:
Looks good to me. Thanks @tomco

PR #1144, opened 2016-03-29
Use bundle-specific plist for Cocoapods resource bundle

feifanzhou:
Resolve https://github.com/Carthage/Carthage/issues/1185

Should restore Carthage and Cocoapods support and app submission


Rupert-RR:
This would certainly be good to have!

grimabe:
+1

PR #1148, opened 2016-03-31
Fix bug where activity indicator wouldn't dismiss

alexandrwang:
To reproduce this bug, send the first message in a conversation,
and the activity indicator will persist forever.

The reason is because the MAX in `messagesAvailableRemotely`
wasn't working properly, and so it wasn't properly capping the
method at 0.


kmahal:
@alexandrwang thank you for this PR.  On of our iOS engineers merged in https://github.com/layerhq/Atlas-iOS/pull/1151 which was a fix for this issue as well before seeing this PR.  Apologies for the mix up and thank you for contributing!

PR #1170, opened 2016-04-19
Adds phone numbers recognizer for chat message

lucacolumbu:
Phone numbers are currently not recognized on the chat messages, visually and on gestures, because NSTextCheckingType was not set to look for phone numbers. I added it for both the UI and the gesture tap on message.


lucacolumbu:
This seems to be actually overwriting the links recognizer, as pointed out by Abir.

PR #1171, opened 2016-04-19
Created protocols and delegate to handle custom input views

kmahal:
This PR allows Atlas to be used with any number of input views for 3rd party library integration. 
1. New `ATLInputView` protocol that all input views would have to implement
2. The new `ATLConversationViewController` delegate `inputViewsForConversationViewController` that takes an array of input views that conform to `ATLInputView`
3. The message input toolbar's left accessory button's action sheet would now have new actionable items defined from the protocol.
4. Atlas handles transitioning from a keyboard to an inputview, and back.

### DIFF
--- a/Code/Controllers/ATLConversationViewController.h
+++ b/Code/Controllers/ATLConversationViewController.h
@@ -32,6 +32,7 @@ typedef NS_ENUM(NSUInteger, ATLAvatarItemDisplayFrequency) {
 
 @class ATLConversationViewController;
 @protocol ATLMessagePresenting;
+@protocol ATLInputView;
 
 ///---------------------------------------
 /// @name Delegate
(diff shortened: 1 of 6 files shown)


Code/Controllers/ATLConversationViewController.h
```diff
@@ -99,6 +100,16 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (NSOrderedSet <LYRMessage*> *)conversationViewController:(ATLConversationViewController *)viewController messagesForMediaAttachments:(NSArray <ATLMediaAttachment*> *)mediaAttachments;
 
+/**
+ @abstract Asks the delegate for an `NSArray` of views that conform to the `ATLInputview` protocol.
```
kcoleman731: Check with @blakewatters  for cocoa protocol naming convention here.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -563,13 +564,40 @@ - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTap
...
+        [self captureLastPhotoTaken];
+    }];
+    [alertController addAction:action2];
+    UIAlertAction *action3 = [UIAlertAction actionWithTitle:ATLLocalizedString(@"atl.conversation.toolbar.actionsheet.library.key", @"Photo/Video Library", nil) style:UIAlertActionStyleDefault handler:^(UIAlertAction * _Nonnull action) {
+        [self showInputView:nil];
+        [self displayImagePickerWithSourceType:UIImagePickerControllerSourceTypePhotoLibrary];
+    }];
+    [alertController addAction:action3];
+    
+    if ([self.delegate respondsToSelector:@selector(inputViewsForConversationViewController:)]) {
```
kcoleman731: Why not check if delegate responds to selector first?

kcoleman731:
What if the developers don't want to show an action sheet? What if they just one a specific thing to happen when the left button is tapped.

Code/Controllers/ATLConversationViewController.m
kmahal: @kcoleman731 either way the 3 actions should be added, and this adds the new actions to the bottom of the action sheet instead of the top as they are optionals

kmahal:
@kcoleman731 this PR doesn't change that.  If they wanted to override the left button action, they can remove all selectors on the button and add their own as they'd do today.  This PR just enhances the already existing functionality of the left accessory button action, not replacing it.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -563,13 +564,40 @@ - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTap
...
+        NSArray *inputViews = [self.delegate inputViewsForConversationViewController:self];
+        for (id<ATLInputView> view in inputViews) {
+            if ([view isKindOfClass:[UIView class]]) {
+                @throw [NSException exceptionWithName:NSInvalidArgumentException reason:@"Input views must be of type `UIView." userInfo:nil];
+            }
+            UIAlertAction *action = [UIAlertAction actionWithTitle:[(id<ATLInputView>)view title] style:UIAlertActionStyleDefault handler:^(UIAlertAction * _Nonnull action) {
```
blakewatters: @kcoleman731 Here `title` needs to be changed to `atlasAlertActionTitle`

kcoleman731: Wow dumb. Thank you.

PR #1172, opened 2016-04-19
Adds phone numbers recognizer for chat message (for real 😀 )

lucacolumbu:
The NSDataDetector was missing the phone number type which prevents to visually identify phone numbers in chat messages(with color and underscore) and it was also preventing the gesture recognizer to trigger the proper action, upon phone number type notification.


chipxsd:
Add spaces around the bitwise OR `|`, and it should be good to merge.

lucacolumbu:
Done!

PR #1174, opened 2016-04-22
Adding logic to switch users to the conversation list view controller

kcoleman731:
* Adds support for user switching.


cashmash:
this was never merged?

PR #1188, opened 2016-05-11
Custom cells for ATLAddresBarViewController.

mailyan:
Added ability to add custom cells for ATLAddresBarViewController. The default cell is ATLParticipantTableViewCell and all the default values are the same.


kcoleman731:
Thanks for the PR @mailyan. I'm going to close your other one and the old issue.

Issue #1070, opened 2015-12-29
#import "ATLAddressBarViewController.h" not found


Alkalouti:
My app support iOS 7
Using Atlas (1.0.11)
Using LayerKit (0.16.0)

Find Attached ..

![untitled](https://cloud.githubusercontent.com/assets/3541656/12035007/b7f6201a-ae40-11e5-87ec-7044616ab7ae.png)

i got this error

mathmatrix828:
+1, using iOS 9 and cocoapods

mathmatrix828:
@Alkalouti, after some research it appears this is an issue where the code is incompatible with Swift 2  - [referenced here](https://github.com/ashleymills/Reachability.swift/issues/21). As per #1074 Swift support has been requested. Not much else to do at the moment, unless you feel like writing the Swift libraries yourself

Rupert-RR:
As a heads up in case people are still having this problem, I came across the same issue and following http://stackoverflow.com/questions/32668723/atladdressbarviewcontroller-h-not-found-in-atlas-h (and some other explanations), added `$(PODS_ROOT)` with the *recursive* option to *Search Paths* in the build settings for my target.

I suspect that this means that it is a CocoaPods integration problem — either CocoaPods should be setting this parameter and it isn't, or the Podspec is not properly configured (I don't know how to do Podspecs).

ksylvest:
Able to reproduce this bug with the following:

`Xcode > File > New > Project...`

```ruby
use_frameworks!
platform :ios, '8.0'

target 'Example' do
  pod 'Atlas'
end
```

```bash
pod --version # 1.0.0.beta.5
pod install
```

Please fix this! It has been broken for almost three months.

feifanzhou:
@ksylvest We are tracking [an issue](https://github.com/CocoaPods/CocoaPods/issues/4950#issuecomment-196289770) with Cocoapods >= 1.0.0b4. That exact setup should install/build fine on Cocoapods 0.39.x.

@Alkalouti @mathmatrix828 Does @Rupert-RR's link fix your build? If not, can you verify your Cocoapods version?

mathmatrix828:
@feifanzhou I've moved away from Atlas at this point, and the project I was planning on using it for didn't pan out. I can fork and do some debugging in my spare time, but unfortunately that has been very limited of late

kmahal:
Closing this issue, but we can re-open if it still exists.

ksylvest:
[...]

PR #1169, opened 2016-04-18
Note that we don't support CocoaPods 1.0.0

feifanzhou:

PR #1191, opened 2016-05-12
[Podspec] Remove header_mappings_dir

mrackwitz:
As all imports assume that the "Code" directory is recursively searched for headers, I'd encourage to not specify a `header_mappings_dir` as this will conserve the directory structure in the header folder within the framework. While a target setup like that itself still compiles fine that causes that the imports of the public headers can't be resolved by the compiler frontend when importing the module as these imports assume a flattened out structure or headers being recursively searched. The latter would need to happen within the framework to avoid Clang complaining about "include of non-modular header inside framework module". That would be generally possible, but isn't implemented by CocoaPods so far.

### DIFF
--- a/Atlas.podspec
+++ b/Atlas.podspec
@@ -17,7 +17,6 @@ Pod::Spec.new do |s|
   s.source_files                = 'Code/**/*.{h,m}'
   s.public_header_files         = 'Code/**/*.h'
   s.ios.resource_bundle         = { 'AtlasResource' => 'Resources/*' }
-  s.header_mappings_dir         = 'Code'
   s.ios.frameworks              = %w{ UIKit CoreLocation MobileCoreServices }
   s.ios.deployment_target       = '8.0'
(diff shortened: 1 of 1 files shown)


mrackwitz:
Note: This would supersede #1169 and fix #1070.

maxeeem:
I updated CocoaPods today and Atlas wouldn't compile complaining about missing headers. This was the key! Thank you @mrackwitz

kcoleman731:
Thanks for this @mrackwitz. I've been trying to sort out the 1.0 podspec issues on my end. 

Have you tried linting the Podspec? I get the following failure. 

screen shot 2016-05-13 at 12 19 43 pm

kcoleman731:
@mrackwitz disregard previous comment...it was a derived data issue with the lint build. Thank you very much for your PR.

mrackwitz:
Always glad to help!

PR #1199, opened 2016-05-23
Using displayName property instead of firstName in address bar.

kcoleman731:


RDorange:
@kcoleman731 
Do you have timeline to package this into a upcoming Cocoapods release ?

No pressure at all, but we have major release mid June migrating from our old messaging system to Layer.com (World Economic Forum, substantial event in China), and just wondering if we should tweak the code ourselves in the meanwhile or not.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Atlas Changelog
 
+## 1.0.24
+
+### Enhancements
+
+* `ATLAddressBarController` now uses the `ATLParticipant` `displayName` property when displaying names in the `ATLAddressBarTextView`.
```
chipxsd: add instead of `firstName`.

PR #1221, opened 2016-06-14
added nullable keyword when needed in ATLConversationViewControllerDe…

gautier-gdx:
Hi, for now I wouldn't be able to use those delegate's methods in swift because I can't make it nil, the xCode won't let me because the nullable keyword wasn't used in it and we needed it in order to reuse default value if needed

Your product is awesome, thx

### DIFF
--- a/Code/Controllers/ATLConversationListViewController.h
+++ b/Code/Controllers/ATLConversationListViewController.h
@@ -135,7 +135,7 @@ NS_ASSUME_NONNULL_BEGIN
  @return A string representing the content of the last message.  If `nil` is returned the controller will fall back to default behavior.
  @discussion This is used when the application uses custom `MIMEType`s and wants to customize how they are displayed.
  */
-- (NSString *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController lastMessageTextForConversation:(LYRConversation *)conversation;
+- (nullable NSString *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController lastMessageTextForConversation:(LYRConversation *)conversation;
 
 /**
  @abstract Asks the data source to configure the query used to fetch content for the controller if necessary.
--- a/Code/Controllers/ATLConversationViewController.h
+++ b/Code/Controllers/ATLConversationViewController.h
@@ -72,7 +72,7 @@ NS_ASSUME_NONNULL_BEGIN
  @discussion Applications should only return a value if the `message` object requires a custom cell class. If `nil` is returned, the collection view will default
  to internal height calculations.
  */
-- (CGFloat)conversationViewController:(ATLConversationViewController *)viewController heightForMessage:(LYRMessage *)message withCellWidth:(CGFloat)cellWidth;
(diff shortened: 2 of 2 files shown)


feifanzhou:
Hey @gautier-gdx,
Thanks for the PR and the kind words!

Currently I won't merge because [`nullable CGFloat`](https://github.com/layerhq/Atlas-iOS/pull/1221/commits/f13cb96e6d45ea63be9e7201932cf779ec85b3d0#diff-28009819fb96130069f3a43019ca7e3eR75) doesn't make much sense. We'd like to move towards standardizing on a value such as `CGFLOAT_MAX` to signal that the collection view should use the default calculations. 

Feel free to update the PR to include that change (and the updated implementation in the collection view), or remove that line from the PR to match the original scope you're going for. Otherwise, I've backlogged the change but it'll be a while before we get a chance to address it.

mlequeux:
/**
 @abstract Asks the delegate for the height to use for a message's cell.
 @param viewController The `ATLConversationViewController` where the message cell will appear.
 @param message The `LYRMessage` object that will be displayed in the cell.
 @param cellWidth The width of the message's cell.
 @return The height needed for the message's cell.
 @discussion Applications should only return a value if the `message` object requires a custom cell class. **If `nil` is returned, the collection view will default
 to internal height calculations.**
 */

You should also change your documentation, please merge this, and add nullable to all the others nullable return values, thanks.

feifanzhou:
@mlequeux I understand your point, but we don't believe that's the right way to resolve this — it's a quick win at the expense of additional tech debt. In particular, it doesn't make sense to make some values `nullable`; `nullable CGFloat` doesn't make sense as `CGFloat` is a primitive.

> We'd like to move towards standardizing on a value such as CGFLOAT_MAX to signal that the collection view should use the default calculations.

mlequeux:
@feifanzhou 

Why are you doing this then ? 
Are you checking for a nullable primitive ? I think 0 should be a valid return value.

```
- (CGSize)heightForMessageAtIndexPath:(NSIndexPath *)indexPath
{
    CGFloat width = self.collectionView.bounds.size.width;
    CGFloat height = 0;
    if ([self.delegate respondsToSelector:@selector(conversationViewController:heightForMessage:withCellWidth:)]) {
        LYRMessage *message = [self.conversationDataSource messageAtCollectionViewIndexPath:indexPath];
        height = [self.delegate conversationViewController:self heightForMessage:message withCellWidth:width];
    }

    //THIS
    if (!height) {
        height = [self defaultCellHeightForItemAtIndexPath:indexPath];
    }

    return CGSizeMake(width, height);
}
```

Otherwise, you should go with NaN to check for invalid CGFloat.

For the method :

`+- (NSOrderedSet <LYRMessage*> *)conversationViewController:(ATLConversationViewController *)viewController messagesForMediaAttachments:(NSArray <ATLMediaAttachment*> *)mediaAttachments;`

Being able to return null is very handful for this one, if you don't want to generate the ATLMediaAttachment for every LYRMessage with a natively handled MIMETYPE.

feifanzhou:
We will likely address these methods one-at-a-time. Conceivable, developers might want to use `0` to indicate a cell that should be fully collapsed; that is more likely than actually wanting a `CGFLOAT_MAX` height cell. The methods that return an object (pointer) can certainly be made `nullable`.

PR #1229, opened 2016-06-28
Resolve compiler errors when integrating library with CocoaPods

jaderfeijo:
# Summary
There are currently two problems when integrating the library with CocoaPods, this Pull Request introduces a couple of changes that solve the issue.

# 1st Issue

* The first problem is that the `LayerKit.framework` file that is included in the `Atlas` target dependencies is pointing to a non-existing directory. Even after a `carthage update` and a `pod install` the location of the framework is different from the dependency target path.

## Solution

* Corrected the `LayerKit.framework` path in the target dependencies to point to the `LayerKit.framework` file that is imported by CocoaPods. This means that, after a `pod install` the project builds without any issues.

# 2nd Issue

* The second problem is that the `Atlas.podspec` specifies a `LayerKit` dependency for versions `0.21.0` or higher. However, the current version of the `Atlas` framework was developed against `0.21.0` and version `0.22.0` of the framework introduces some breaking changes to the `LayerKit` framework API (against semantic versioning conventions). This loose dependency version specification causes the `LayerKit` framework version `0.22.0` (latest version as of the date of this PR) to be mistakenly installed on target projects when running `pod install`, which causes another build error due to the breaking API changes in version `0.22.0`.

## Solution

* The solution was to fix the required library version to `0.21.0` in the `Atlas.podspec` file to ensure the correct library version is imported into consuming projects.

### DIFF
--- a/Atlas.xcodeproj/project.pbxproj
+++ b/Atlas.xcodeproj/project.pbxproj
@@ -7,7 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		250C5EF31C7656610084EB18 /* LayerKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 250C5EF21C7656610084EB18 /* LayerKit.framework */; };
 		251C57AD1C5A9E3C00110871 /* ATLBaseCollectionViewCell.m in Sources */ = {isa = PBXBuildFile; fileRef = 251C57AC1C5A9E3C00110871 /* ATLBaseCollectionViewCell.m */; };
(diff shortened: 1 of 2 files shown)


jaderfeijo:
I believe the CI has failed due to an old version of `cocoapods` being used by the CI system. The line that is causing integration to fail was not actually introduced by the PR.

screen shot 2016-06-29 at 09 31 08

feifanzhou:
@jaderfeijo thanks for this PR! First part looks good. We're in the process of updating Atlas for LayerKit 0.22.0 … if you could isolate the first commit, I'll merge this

jaderfeijo:
@feifanzhou sure thing.

jaderfeijo:
@feifanzhou all done :)

PR #1232, opened 2016-06-30
Fix for handling avatar visibility when using shouldDisplayAvatarItem

daniel-maness:
Exposes the shouldDisplayAvatar property of an ATLBaseCollectionViewCell.  This fixes the shouldDisplayAvatarItem method by allowing it to set the visibility of the cell's ATLAvatarImageView even if a message has no sender.  Previously only updateWithSender was setting this, but not shouldDisplayAvatarItem.


jpmcglone:
I think we should close this @daniel-maness because of soon changes to AvatarView for Presence. If this is important, we can revisit. What are your thoughts?

daniel-maness:
Yep that sounds good, closing.

Issue #1216, opened 2016-06-09
ATLConversationListViewControllerDataSource lastMessageTextForConversation should return nullable NSString


V-FEXrt:
I was following this tutorial https://developer.layer.com/docs/ios/atlas#configure-conversationlistviewcontroller and at the end of the function it returns nil. I am writing my app with Swift, and without the nullable keyword a nil value cannot be returned.

Changing 

``` ObjC
- (NSString *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController lastMessageTextForConversation:(LYRConversation *)conversation;
```

to

``` ObjC
- (nullable NSString *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController lastMessageTextForConversation:(LYRConversation *)conversation;
```

in ATLConversationListViewController.h fixed my issue.

I tested adding the nullable keyword and that fixed the issue. This is likely also the case for ConversationViewControllerDataSource messagesForMediaAttachments as I am also unable to return nil as shown in https://developer.layer.com/docs/ios/atlas#configure-conversationviewcontroller part 4

V-FEXrt:
Resovled by #1341

PR #1246, opened 2016-08-04
Add nullable return type for functions with 'nil' based functionality. Addresses issue #1216

V-FEXrt:
These two functions have functionality dependent on nil being returned, so nullable must be added for Swift


cpiggott:
Can anyone update why this failed? It seems like most of the recent travis-ci tests have failed.

PR #1255, opened 2016-09-13
Bugfix/atlas query controller setup

daniel-maness:
The ATLConversationListViewController query controller should only be initialized once within viewWillAppear.


oferRounds:
Following this change I sometimes get the following crash on ATLConversationListViewController.m, where on the method: setupConversationQueryController this assertion fails:
    `NSAssert(self.queryController == nil, @"Cannot initialize more than once");`

PR #1265, opened 2016-09-27
[WIP] Bumping LayerKit (Update Carthage)

LukeDurrant:
Fix Carthage support 
LYRIdentity model was missing. which is because Carthage was pointing to an older version of LayerKit (v0.19.1) 

See https://github.com/Carthage/Carthage/issues/1308


LukeDurrant:
Someone will need to publish the pod spec before the tests will pass

Issue #1251, opened 2016-09-07
Feature request: options for customising the date format in the conversation list


elitalon:
Hello,

Currently it is possible to customise the date of the section headers inside `ATLConversationViewController` through the data source:

``` objc
- (NSAttributedString *)conversationViewController:(ATLConversationViewController *)conversationViewController attributedStringForDisplayOfDate:(NSDate *)date
```

It would be nice to offer a customisation point also for the date of the last message in the conversation list, for instance through the data source:

``` objc
- (NSAttributedString *)conversationListViewController:(ATLConversationListViewController *)conversationListViewController attributedStringForDisplayOfDate:(NSDate *)date
```

Another possibility would be a property, since the format should ideally be the same in every cell:

``` obj
@property (nonatomic, strong, nonnull) NSDateFormatter *lastMessageDateFormatter;
```

Are there any plans in this direction?

Thank you!

PR #1275, opened 2016-09-29
Add delegate method for custom dates in conversation list

elitalon:
This addresses #1251 by providing an optional method in `ATLConversationListViewControllerDataSource`, which can be used to feed a custom date.

Otherwise it fall backs to the default behaviour, which has been extracted from `ATLConversationTableViewCell`.


elitalon:
I cannot run the tests due to an issue with KIF. I have reported it and [submitted a PR](https://github.com/blakewatters/KIFViewControllerActions/pull/6) for that too.

PR #1279, opened 2016-10-03
Moved notification observers from  to  to avoid race condition

daniel-maness:


Code/Controllers/ATLConversationListViewController.m
```diff
@@ -171,6 +166,11 @@ - (void)viewWillAppear:(BOOL)animated
             [self.tableView selectRowAtIndexPath:selectedIndexPath animated:NO scrollPosition:UITableViewScrollPositionNone];
         }];
     }
+    
+    // Track changes in authentication state to manipulate the query controller appropriately
```
chipxsd: While I see the concern of the race condition, `viewWillAppear:` can get called multiple times (say you're either transitioning or just popping up a modal view controller on top of this one), this will make the controller register to notifications more than once. Unlike `viewDidLoad` which is invoked only once.

If you really need to register it on `viewWillAppear:` then make sure you remove the observer on `viewWillDissapear:` as well.

chipxsd: I think it might be cleaner registering them and removing them `viewWillAppear:` and `viewWillDissapear:` respectively, this way `NSNotificationCenter` has less work to do when the view is not in focus.

daniel-maness: Cool I like that. Changes pushed.

feifanzhou:
Looks good

PR #1280, opened 2016-10-04
Removed query controller delegate when a conversation is removed from…

daniel-maness:
… ATLConversationViewController. This fixes the following case:

1. Start a new conversation
2. Add a participant for an existing conversation so the collectionView loads those messages
3. Remove all participants
4. Send, from another device, a message to the conversation that was just unloaded
5. NSInternalInconsistencyException is thrown


feifanzhou:
Looks good

PR #1283, opened 2016-10-07
APPS-2524 Conversation Pagination (MAK)

markkrenek:
Adds pagination (using LYRQueryController.paginationWindow, set to 30). Pulls in the next page of conversations when scrolled close to the bottom of the table view. Activity spinner placed in table view footer while there are additional pages of conversations.

NOTE: There is another branch (APPS-2524-conversation-pagination) that attempts the same functionality. I created by own branch (signified with my initials MAK in the name) instead of using that one.

### DIFF
--- a/Code/Controllers/ATLConversationListViewController.m
+++ b/Code/Controllers/ATLConversationListViewController.m
@@ -559,7 +559,7 @@ NSString *const ATLConversationListViewControllerDeletionModeEveryone = @"Everyo
     }
 }
 
-- (void) expandPaginationWindow
+- (void)expandPaginationWindow
 {
     self.queryController.paginationWindow += self.queryController.paginationWindow + ATLConverstionListPaginationWindow < self.queryController.totalNumberOfObjects ? ATLConverstionListPaginationWindow : self.queryController.totalNumberOfObjects - self.queryController.paginationWindow;
 }
--- a/Code/Controllers/ATLConversationListViewController.m
+++ b/Code/Controllers/ATLConversationListViewController.m
@@ -27,6 +27,7 @@ static NSString *const ATLImageMIMETypePlaceholderText = @"Attachment: Image";
 static NSString *const ATLVideoMIMETypePlaceholderText = @"Attachment: Video";
 static NSString *const ATLLocationMIMETypePlaceholderText = @"Attachment: Location";
 static NSString *const ATLGIFMIMETypePlaceholderText = @"Attachment: GIF";
+static NSInteger const ATLConverstionListPaginationWindow = 30;
 
 @interface ATLConversationListViewController () <UIActionSheetDelegate, LYRQueryControllerDelegate, UISearchBarDelegate, UISearchControllerDelegate, UISearchDisplayDelegate>
 
@@ -261,6 +262,7 @@ NSString *const ATLConversationListViewControllerDeletionModeEveryone = @"Everyo
         NSLog(@"LayerKit failed to create a query controller with error: %@", error);
         return;
     }
+    self.queryController.paginationWindow = ATLConverstionListPaginationWindow;
     self.queryController.delegate = self;
     
     BOOL success = [self.queryController execute:&error];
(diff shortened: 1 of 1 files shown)


Code/Controllers/ATLConversationListViewController.m
```diff
@@ -508,6 +513,9 @@ - (void)queryController:(LYRQueryController *)controller
 - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
 {
     [self.tableView endUpdates];
+
+    [self configureMoreConversationsFooterView];
```
chipxsd: rename to
```objective-c
- (void)configureLoadingMoreConversationsIndicatorView;
```

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -517,7 +525,81 @@ - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
...
+    if (decelerate) return;
+    [self configurePaginationWindow];
+}
+
+- (void)scrollViewDidEndDecelerating:(UIScrollView *)scrollView
+{
+    [self configurePaginationWindow];
+}
+
+- (void)scrollViewDidScrollToTop:(UIScrollView *)scrollView
```
chipxsd: I haven't tried the code though, but is this one necessary?

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -517,7 +525,81 @@ - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
...
+}
+
+- (BOOL)moreConversationsAvailable
+{
+    return self.queryController.paginationWindow < self.queryController.totalNumberOfObjects;
+}
+
+- (BOOL)isNearBottom
+{
+    CGFloat minimumDistanceFromBottomToTriggerLoadingMore = 200;
```
chipxsd: Define the `200` as a const `CGFloat` at the top of the implementation file:
```objective-c
static CGFloat const ATLConversationListLoadMoreConversationsDistanceThreshold = 200.0f; // or trigger, whichever you prefer
```

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -517,7 +525,81 @@ - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
...
+
+- (void)configureMoreConversationsFooterView
+{
+    BOOL moreConversationsAvailable = [self moreConversationsAvailable];
+    if (moreConversationsAvailable == self.showingMoreConversationsIndicator) return;
+    self.showingMoreConversationsIndicator = moreConversationsAvailable;
+
+    // This footer view holds an activity indicator view. When there are more conversations to load, the height of the view is set to expose the activity indicator. Otherwise, the height is set to zero, but the footer remains installed in the table view. This is required in order to suppress the dummy separator lines that UITableView draws to simulate empty rows.
+    UIView* footer = self.tableView.tableFooterView;
+    footer.frame = self.showingMoreConversationsIndicator ? CGRectMake(0, 0, 0, 30) : CGRectZero;
```
chipxsd: Define the `30` as a const `CGFloat` at the top of the implementation file:
```objective-c
static CGFloat const ATLConversationListViewLoadMoreConversationIndicatorViewFrameHeight = 30.0f;
```

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -517,7 +525,81 @@ - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
...
+
+    // Must reassign tableFooterView in order for UITableView to recognize the size change.
+    self.tableView.tableFooterView = footer;
```
chipxsd: Maybe there's a method that does that for you, I haven't done UIKit in a while, but try a few:
```objetive-c
[self.tableView.tableFooterView sizeToFit];
[self.tableView.tableFooterView setNeedsLayout];
[self.tableView sizeToFit];
```

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -517,7 +525,81 @@ - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
...
+
+- (UIView*)makeMoreConversationsFooterView
+{
+    UIView* view = [[UIView alloc] init];
+
+    UIActivityIndicatorView* activityIndicatorView = [[UIActivityIndicatorView alloc] init];
+    activityIndicatorView.autoresizingMask = UIViewAutoresizingFlexibleLeftMargin | UIViewAutoresizingFlexibleRightMargin | UIViewAutoresizingFlexibleTopMargin | UIViewAutoresizingFlexibleBottomMargin;
+    activityIndicatorView.activityIndicatorViewStyle = UIActivityIndicatorViewStyleGray;
+    [activityIndicatorView startAnimating];
+    [view addSubview:activityIndicatorView];
```
chipxsd: Also not sure, but you can give it a try, see if you can return the `UIActivityIndicatorView` directly (and apply the autoresizingMask same as you're doing it to the `UIView`).

markkrenek:
I'll make these changes. Some of the coding style came from the previous branch that I was referencing when making this change.

Code/Controllers/ATLConversationListViewController.m
```diff
@@ -517,7 +525,81 @@ - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
     }
 }
 
+#pragma mark - UIScrollViewDelegate
+
+- (void)scrollViewDidEndDragging:(UIScrollView *)scrollView willDecelerate:(BOOL)decelerate
+{
+    if (decelerate) return;
```
markkrenek: Should I bring the rest of the file up to this convention? There are several if blocks that don't use curly braces and new lines.

Code/Controllers/ATLConversationListViewController.m
markkrenek: UIActivityIndicatorView size is generally controlled by the OS. It does seem however in this case, I can force it's height to 30 when installed in as the table view's footer. Normally it's height is 20.

PR #1291, opened 2016-10-14
fixes alignment constraint never removed so avatar is almost full width

geoffmacd:
When calling -configureCellForType, the layout can incorrectly apply a near full width avatarImageView because the (12) padding on the left/right depending on the previous type was never removed. This PR persists and removes that constraint when the cellType is dynamically changed. This was not noticed if the cellType was never changed as is the case using ATLIncomingMessageCollectionViewCell / Outcoming as they are only set once.

This is what happened previously:
simulator screen shot oct 14 2016 11 02 33 am

Notice the 12-avatar-12 constraints which were left

screen shot 2016-10-14 at 11 08 55 am


geoffmacd:
I dont know how people could have implemented custom cells without this being a problem

PR #1295, opened 2016-10-17
Bugfix/collection view cell misaligned

daniel-maness:
Fixes avatar alignment constraint never being removed, resulting in almost full width avatar images.


geoffmacd:
this is an easy win

PatrickNLT:
+1.
I just encountered this bug and made the exact same fix.

PR #1303, opened 2016-10-28
Show time in messages and date in sections if day changed

bimusiek:
Hey, few people asked about this feature and we also needed it so I coded time in messages. 

If you feel like it could be mergeable after some steps, please tell me what's left to be done from my side. 

Cheers, Mike

How it looks: 
screenshot 2016-10-28 12 00 22


fmecklen:
We also need this feature. This would be amazing

PR #1308, opened 2016-11-03
Fix rightAccessoryButton with image being forced to text

rivera-ernesto:
It was impossible to show an image in the button when the text is not empty as `configureRightAccessoryButtonForText` was called no matter what.


Code/Views/ATLMessageInputToolbar.m
```diff
@@ -370,17 +370,12 @@ - (void)acceptAutoCorrectionSuggestion
 
 - (void)configureRightAccessoryButtonState
 {
-    if (self.textInputView.text.length) {
-        [self configureRightAccessoryButtonForText];
```
rivera-ernesto: This removes the image from the button as soon as some text is input.

PR #1316, opened 2016-11-18
Add appearance setter for left participant padding

tschuls:
Since we do not use the Avatar icon, we would like to change the default left padding of the participant label to a smaller value.
With the presented changes the left padding is customizable.


delox:
Merge this already!

PR #1330, opened 2016-12-09
Changed queryController delegates to not fire if the view is not visi…

daniel-maness:
…ble. This should avoid some instances of the common NSInternalInconsistencyException


Code/Controllers/ATLConversationViewController.m
```diff
@@ -1222,6 +1224,12 @@ - (void)queryControllerWillChangeContent:(LYRQueryController *)queryController
 
 - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
 {
+    if (self.collectionView.window == nil) {
+        [self.collectionView reloadData];
+        [self.collectionView layoutIfNeeded];
+        return;
+    }
+
```
geoffmacd: I think you should put this block after the removal of objectChanges just in case -didChangeObject: is called just before a window is remove and you have a leftover changes that were already applied by reloadData

daniel-maness: Good call. Changed it so `objectChanges` is always removed.

PR #1350, opened 2017-02-08
Refactored to track all changes from the queryController as they come…

daniel-maness:
… in instead of modifying the tableView. Then, when the queryController finishes with its changes, update the tableView with all changes at once.


Code/Controllers/ATLConversationListViewController.m
```diff
@@ -536,6 +541,30 @@ - (void)queryControllerDidChangeContent:(LYRQueryController *)queryController
     }
 }
 
+- (NSMutableArray *)insertedRowIndexPaths
+{
+    if (_insertedRowIndexPaths == nil) {
+        _insertedRowIndexPaths = [[NSMutableArray alloc] init];
+    }
+    return _insertedRowIndexPaths;
```
jpmcglone: how do you feel about 
```
return _insertedRowIndexPath ?: (_insertedRowIndexPaths = [[NSMutableArray alloc] init]);
```

daniel-maness: Yeah much cleaner, I like it. Will update.

jwyld:
This is essentially what we did with Cola, but we abstracted to a "change set" object which held all the changes.  We were able to reuse this for the conversation as well as the conversation list.  In Atlas, the code has gone through a number of iterations along this similar code path.  I'm not sure if we're seeing failures there.  If not, I would ask why not.  If we are seeing crashes, we should apply similar treatment there as well.

LukeDurrant:
Nice!

PR #1356, opened 2017-02-21
Added fastlane support for testing Atlas

jpmcglone:


fastlane/README.md
```diff
@@ -0,0 +1,42 @@
+fastlane documentation
```
chipxsd: Can we put this guys into `.gitignore`, in fact everything that gets autogenerated should be git ignored.

.gitignore
```diff
@@ -1,10 +1,204 @@
-.DS_Store
-xcuserdata
-*.framework
-*.bundle
 Pods
-*.xccheckout
 Releases
 xcodebuild.log
 
+# Created by https://www.gitignore.io/api/xcode,swift,fastlane,carthage,appcode,ruby
```
chipxsd: Can you clean up the git ignore file? Remove instructions and everything that's disabled, you can leave the section comments.

jpmcglone: Removing commented out stuff. Leaving section comments (they're useful actually haha). Good idea.

fastlane/README.md
jpmcglone: I found `skip_docs`. Going to do that instead. If they ever make better README's we can remove that.

.gitignore
chipxsd: No, so the `.gitignore` file you added was computer generated with instructions for **you**, not for me, not for our customers, but only for **you**, **you** are checking in that file into the repo, which makes **you** responsible for knowing which rule to add to the list and which ones to take out.

In this commit, I don't have a clear sight to which rules are taking effect, because there's a bunch of leftover instructions made by **someone** (which is the author of https://www.gitignore.io) for **someone** (you) who's responsibility is to comb through the list, make changes and clean it up, and check-in a clean `.gitignore` file.

The `.gitignore` file should have rules in them and comments describing the reasoning behind the rules that were added (if it's not self explanatory).

This to me looks like leftovers from a template (and to anyone who'll open the `.gitignore` file):
```
# for a library or gem, you might want to ignore these files since the code is
# intended to run in multiple environments; otherwise, check them in:
# Gemfile.lock
# .ruby-version
# .ruby-gemset
```

Again, these comments are not for our customers, nor for me since I'm not the one adding rules to the `.gitignore` file.

jpmcglone: done.

jpmcglone:
Will merge when Travis passes

PR #1360, opened 2017-02-24
use master spec repo version of KIFViewControllerActions

armcknight:


armcknight:
yeah we were only using that because we hadn't pushed 1.0.2 to pod trunk

PR #1374, opened 2017-04-07
APPS-2740 - Presence Feature

jpmcglone:


Code/Atlas.h
```diff
@@ -90,6 +90,7 @@
 #import "ATLProgressView.h"
 #import "ATLUIImageHelper.h"
 #import "ATLPlayView.h"
+#import "ATLPresenceStatusView.h"
```
armcknight: Purely preferential, I always sort imports. Here's an automator service (Automator -> `⌘N` -> Service) I have that sorts all selected lines. Then you can create a custom shortcut for it, I have it mapped to `⌘⌥^S`
screen shot 2017-04-21 at 2 19 44 pm

Code/Views/ATLAvatarView.h
```diff
@@ -59,6 +59,16 @@ extern CGFloat const ATLAvatarImageDiameter;
 @property (nonatomic) UIColor *imageViewBackgroundColor UI_APPEARANCE_SELECTOR;
 
 /**
+ @abstract Sets whether the presence status is shown for this AvatarView. Default is true.
+ */
+@property (nonatomic) BOOL presenceStatusEnabled;
+
+/**
+ @abstract Sets the presenceStatus for the AvatarView, which automatically styles the ATLPresenceStatusView
```
armcknight: If we're gonna put in a capitalized camelcased `AvatarView`, may as well add the `ATL` prefix.

Code/Views/ATLAvatarView.m
```diff
@@ -71,13 +67,17 @@ - (void)lyr_commonInit
...
     _initialsFont = [UIFont systemFontOfSize:14];
     _initialsColor = [UIColor blackColor];
-    _avatarImageViewDiameter = 27;
-    
-    self.clipsToBounds = YES;
-    self.layer.cornerRadius = _avatarImageViewDiameter / 2;
+
     self.contentMode = UIViewContentModeScaleAspectFill;
-    self.accessibilityLabel = ATLAvatarImageViewAccessibilityLabel;
+    self.accessibilityLabel = ATLAvatarViewAccessibilityLabel;
```
armcknight: Unless there's an overridden setter for `accessibilityLabel`, I suggest replacing `self.` with `_` for consistency and also I believe that is best practice for `@property` initialization in objc.

Code/Views/ATLAvatarView.m
```diff
@@ -134,18 +140,53 @@ - (void)setInitialsFont:(UIFont *)initialsFont
...
+        case LYRIdentityPresenceStatusBusy:
+            _presenceStatusView.statusColor = [UIColor colorWithRed:230.0/255.0 green:68.0/255.0 blue:63.0/255.0 alpha:1.0];
+            _presenceStatusView.mode = ATLMPresenceStatusViewModeFill;
+            break;
+        case LYRIdentityPresenceStatusAway:
+            _presenceStatusView.statusColor = [UIColor colorWithRed:247.0/255.0 green:202.0/255.0 blue:64.0/255.0 alpha:1.0];
+            _presenceStatusView.mode = ATLMPresenceStatusViewModeFill;
+            break;
+        case LYRIdentityPresenceStatusInvisible:
+            _presenceStatusView.statusColor = [UIColor colorWithRed:79.0/255.0 green:191.0/255.0 blue:98.0/255.0 alpha:1.0];
```
armcknight: Could we define these colors in a reusable, self documenting way?

There are many ways I've done this in the past:
- category on UIColor, less preferred here because I don't like delivering categories in SDKs (requires the -ObjC linker flag in consumer projects)
- `#defines`
- `FOUNDATION_EXPORT`ed constants

Don't really have a preference, other than not using "magic colors".

Code/Views/ATLAvatarView.m
```diff
@@ -188,18 +229,34 @@ - (void)updateWithImage:(UIImage *)image forRemoteImageURL:(NSURL *)remoteImageU
...
-    [self addConstraint:[NSLayoutConstraint constraintWithItem:self.initialsLabel attribute:NSLayoutAttributeLeft relatedBy:NSLayoutRelationEqual toItem:self attribute:NSLayoutAttributeLeft multiplier:1.0 constant:3]];
-    [self addConstraint:[NSLayoutConstraint constraintWithItem:self.initialsLabel attribute:NSLayoutAttributeRight relatedBy:NSLayoutRelationEqual toItem:self attribute:NSLayoutAttributeRight multiplier:1.0 constant:-3]];
-    [self addConstraint:[NSLayoutConstraint constraintWithItem:self.initialsLabel attribute:NSLayoutAttributeTop relatedBy:NSLayoutRelationEqual toItem:self attribute:NSLayoutAttributeTop multiplier:1.0 constant:3]];
-    [self addConstraint:[NSLayoutConstraint constraintWithItem:self.initialsLabel attribute:NSLayoutAttributeBottom relatedBy:NSLayoutRelationEqual toItem:self attribute:NSLayoutAttributeBottom multiplier:1.0 constant:-3]];
+    [super layoutSubviews];
+    
+    // Image View
+    self.imageView.frame = self.bounds;
+    
+    CGFloat avatarViewDiamater = MIN(self.imageView.bounds.size.width, self.imageView.bounds.size.height);
```
armcknight: Never reach into the bounds struct directly in ObjC, use these functions:
- CGRectGetWIdth/CGRectGetWidth for size values
- CGRectGetMinX/CGRectGetMinY for origin coordinates (there are also mid/max variants)

Code/Views/ATLParticipantTableViewCell.m
```diff
@@ -169,14 +168,14 @@ - (void)configureNameLabelConstraints
     
     // NOTE: We're not using NSLayoutRelationLessThanOrEqual here because doing so would cause iOS 8.0 to not update the label's intrinsic content size constraints when the label's value is changed / the cell is reused.
```
armcknight: I'm confused as to why this comment is here. Could someone add to it the explanation for why `NSLayoutRelationLessThanOrEqual` was considered (but rejected)?

Code/Views/ATLPresenceStatusView.h
```diff
@@ -0,0 +1,49 @@
+//
+//  ATLPresenceStatusView.h
+//  Atlas
+//
+//  Created by JP McGlone on 4/5/17.
+//  Copyright (c) 2015 Layer, Inc. All rights reserved.
```
armcknight: update copyright year

Code/Views/ATLPresenceStatusView.h
```diff
@@ -0,0 +1,49 @@
+//
+//  ATLPresenceStatusView.h
+//  Atlas
+//
+//  Created by JP McGlone on 4/5/17.
```
armcknight: Open source projects I've worked on typically omit author info, not sure on our rules.

Code/Views/ATLPresenceStatusView.h
```diff
@@ -0,0 +1,49 @@
...
+    ATLMPresenceStatusViewModeFill,
+    ATLMPresenceStatusViewModeBordered
+};
+
+@interface ATLPresenceStatusView : UIView
+
+/**
+ @abstract Sets the color of the presence status fill or border. Default is [UIColor lightGrayColor]
+ */
+@property (nonatomic) UIColor *statusColor;
```
armcknight: Are we purposefully not making this a `UI_APPEARANCE_SELECTOR`? (Sorry if this has been discussed.) I see that it's passed in the init, why do we use two different approaches for `statusColor` and `statusBackgroundColor`?

Code/Views/ATLPresenceStatusView.m
```diff
@@ -0,0 +1,116 @@
...
+        _mode = mode;
+    }
+    return self;
+}
+
+- (instancetype)initWithFrame:(CGRect)frame
+{
+    self = [super initWithFrame:frame];
+    if (self) {
+        self.backgroundColor = [UIColor clearColor];
```
armcknight: I believe this is the default value for `backgroundColor`, so we could omit this line.

Code/Views/ATLPresenceStatusView.m
```diff
@@ -0,0 +1,116 @@
...
+    CGFloat radius = diameter * 0.5;
+    
+    // Draw background first
+    CGContextSetFillColorWithColor(context, _statusBackgroundColor.CGColor);
+    CGContextAddArc(context, center.x, center.y, radius, 0.0, M_PI*2, YES);
+    CGContextFillPath(context);
+    CGContextSaveGState(context);
+
+    // Set width to 1/6 the circle width
+    CGFloat borderWidth = diameter * 0.166;
```
armcknight: The comment is pretty redundant. Why not 
```
CGFloat borderWidth = diameter / 6.0;
```
?

Code/Views/ATLPresenceStatusView.m
```diff
@@ -0,0 +1,116 @@
...
+            // Fill the circle
+            CGContextSetFillColorWithColor(context, _statusColor.CGColor);
+            CGContextAddArc(context, center.x, center.y, radius - (borderWidth * 0.5), 0.0, M_PI*2, YES);
+            CGContextFillPath(context);
+            break;
+        }
+        case ATLMPresenceStatusViewModeBordered:
+        {
+            // Inset the radius
+            CGContextAddArc(context, center.x, center.y, radius - borderWidth, 0.0, M_PI*2, YES);
```
armcknight: Extract the magic numbers in this file to constants.

README.md
```diff
@@ -268,13 +268,24 @@ Atlas takes advantage of Apple's `UIAppearance` protocol which lets you change U
...
+### [ATLAvatarView](Code/Views/ATLAvatarView.h)
 
 ```objc
+@property (nonatomic) UIImage *imageView;
 @property (nonatomic) CGFloat avatarImageViewDiameter
 @property (nonatomic) UIFont *initialsFont
 @property (nonatomic) UIColor *initialsColor
 @property (nonatomic) UIColor *imageViewBackgroundColor
+@property (nonatomic) BOOL presenceStatusEnabled;
+@property (nonatomic) LYRIdentityPresenceStatus presenceStatus;
```
armcknight: Why do we replicate the interfaces in the Readme? That's a lot of extra maintenance burden to keep in sync. The `.h` files should be the one and only source of truth.

Code/Views/ATLPresenceStatusView.h
jpmcglone: 1. if we were it wouldn't go here, really. We'd have to have colors per presence or something, to make it sensible.

2. AFAIK no, we're doing one set of colors across the platforms right now.

3. Background color is the circle behind the status. Status color is the status. So e.g. in most places, the bg color is white. I think this is a good first pass. Ideally, perhaps we get rid of background color (or at least make it "clear" for default) and use it to 'clip' the view underneath it. This requires more discussion, so I did only what the task asked of us (and I think this aligns with how Android is doing it)

Code/Views/ATLPresenceStatusView.m
jpmcglone: +1 if true

Code/Views/ATLPresenceStatusView.m
jpmcglone: Idk if true, but at a past job, one of our engineers said it was more performant to multiply by a float than divide. Regardless, our code base never used divides when we could have easily multiplied. So for me, its just an old habit. I really don't care if this gets updated or not

Code/Views/ATLAvatarView.m
```diff
@@ -87,18 +87,22 @@ - (void)lyr_commonInit
...
+    // Presence Status View
+    _presenceStatusView = [[ATLPresenceStatusView alloc] init];
+    _presenceStatusEnabled = true;
+    [self addSubview:_presenceStatusView];
 }
 
 - (CGSize)intrinsicContentSize
 {
-    return CGSizeMake(self.avatarImageViewDiameter, self.avatarImageViewDiameter);
+    return CGSizeMake(30, 30);
```
chipxsd: I'm not sure I get this line, I don't see nothing wrong with keeping it the old way, since it's configurable via the public property.

Code/Views/ATLPresenceStatusView.m
```diff
@@ -0,0 +1,116 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+#import "ATLPresenceStatusView.h"
+
+@import CoreGraphics;
```
chipxsd: what's our policy now with the `@import` vs `#import`?

chipxsd:
Also, do we have any policy around versioning (semantic), there's a breaking change `ATLAvatarImageView` got renamed to `ATLAvatarView` along with some properties.

Code/Views/ATLAvatarView.m
chipxsd: you can't do it on subclassed properties

Code/Views/ATLPresenceStatusView.m
chipxsd: well, true that cpu mul/mod instructions cost different, but let's leave that to the compiler to optimize. Plus you lose a lot of precision, if you invert the float number, especially rounded at the third digit.

jwyld: hahaha... all that worry over clock performance of multiply versus divide yet `.bound` is executed 4 times!

Code/Views/ATLPresenceStatusView.m
jwyld: I thought JP removed a bunch of these.  I see a lot more of them in this PR.

Code/Views/ATLPresenceStatusView.m
daniel-maness: Default value for `backgroundColor` is `nil`, and a quick test by removing it turns the background black in the UI, so it's needed.

Code/Views/ATLPresenceStatusView.m
chipxsd: I'll fix these

Code/Views/ATLPresenceStatusView.h
daniel-maness: The rest of Atlas is filled with author info right now. Probably @blakewatters call on what the rule is going forward

blakewatters: I don't particularly care right this second. If we want to make any changes to enforce a standard header that's fine but it needs automation to ensure conformance

Code/Views/ATLAvatarView.m
daniel-maness: @chipxsd `avatarImageViewDiameter` was removed from `ATLAvatarView`, and trying to use the size of `self.avatarItem.imageView` fails because it's `nil` at this point, so nothing shows for the avatar. I'll do some refactoring to get it working without using magic numbers.

PR #1375, opened 2017-04-10
Updates to Atlas Project that make it easier to test locally with

jpmcglone:
- Removed `Atlas` source from `Atlas` project. To access, use the Pods -> Development Pods -> Atlas
- Added the `Layer` source back to the Podfile
- Updated `AtlasHost` scheme to use the Pods' `Atlas` (some reason, this is required for any of the tests that access the photo library. It works, and looks identical to what is currently in master)
- Updated project settings via xcode suggestion


Podfile
```diff
@@ -4,9 +4,8 @@ platform :ios, '8.0'
 use_frameworks!
 
 if !ENV['LAYER_USE_CORE_SDK_LOCATION'].blank? then
+  source 'git@github.com:layerhq/cocoapods-specs.git'
```
chipxsd: wait, why did this get back in here??

jpmcglone: needed for LayerThrift dependency in LayerKit

jpmcglone: *only* needed when building LayerKit locally.

chipxsd: @armcknight and I already had this conversation on #ios channel already.

To recap, this environment variable `LAYER_USE_CORE_SDK_LOCATION` is great not only for us (developers at @layerhq/developers) to point at the LayerKit sources, but also for developers pointing to any arbitrary pre-compiled version of LayerKit they want to try out.

Now we're literally exposing, again, one of our private repos. If anybody outside layer uses the `LAYER_USE_CORE_SDK_LOCATION` environment variable, they'll trigger this condition in ruby that will make `pod` fetch an unknown CocoaPods repo that **nobody** has access to, and throw a stack trace.

screen shot 2017-04-10 at 4 07 39 pm

PR #1377, opened 2017-04-12
Adjust ATLMaxCellWidth for wider devices

rivera-ernesto:
Adds the extra width points available when compared to a 320p-wide device


rivera-ernesto:
Yeah. I think that making it wider is important for wide devices, maybe just not growing linearly.

Maybe something like `return max(215, screenWidth * 0.7)`?

jwyld:
After a lot of testing with Messages, I found 0.72 seems to be about the correct value.

rivera-ernesto:
Right. Let me make those changes.

PR #1378, opened 2017-04-17
Fixes timestamp overlay bug

### LINKED ISSUE #1067: Timestamp appears ontop of message bubble
I only notice this when the chat controller gets updated with multiple messages. For example, coming in and out of network and you have queued messages to be displayed. Happens on the update. 

photo

jpmcglone:
resolve #1067

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -483,9 +483,14 @@ static NSInteger const ATLPhotoActionSheet = 1000;
     header.message = message;
     if ([self shouldDisplayDateLabelForSection:indexPath.section]) {
         [header updateWithAttributedStringForDate:[self attributedStringForMessageDate:message]];
+    } else {
+        [header updateWithAttributedStringForDate:[[NSAttributedString alloc] initWithString:@"" attributes:nil]];
     }
+    
     if ([self shouldDisplaySenderLabelForSection:indexPath.section]) {
         [header updateWithParticipantName:[self participantNameForMessage:message]];
+    } else {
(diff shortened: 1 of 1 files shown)


Code/Controllers/ATLConversationViewController.m
```diff
@@ -483,9 +483,14 @@ - (void)configureHeader:(ATLConversationCollectionViewHeader *)header atIndexPat
...
     if ([self shouldDisplayDateLabelForSection:indexPath.section]) {
         [header updateWithAttributedStringForDate:[self attributedStringForMessageDate:message]];
+    } else {
+        [header updateWithAttributedStringForDate:[[NSAttributedString alloc] initWithString:@"" attributes:nil]];
     }
+    
     if ([self shouldDisplaySenderLabelForSection:indexPath.section]) {
         [header updateWithParticipantName:[self participantNameForMessage:message]];
+    } else {
+        [header updateWithAttributedStringForDate:[[NSAttributedString alloc] initWithString:@"" attributes:nil]];
```
prabhupb: Shouldn't this be ` [header updateWithParticipantName:@""];` ?

prabhupb: Currently this hides all the timestamps if the sender label is not displayed.

edwinruan: Is that a way to display timestamp but hide sender Label? Look like hiding both is just a workaround.

edwinruan: Is that a way to display timestamp but hide sender Label? Look like hiding both is just a workaround.

vinzen:
#1404

Code/Controllers/ATLConversationViewController.m
vinzen: PR created #1404

PR #1379, opened 2017-04-21
APPS-2740 - Presence Feature

chipxsd:
Adds the new presence feature introduced with LayerKit v0.26.0.


chipxsd:
I don't get this PR. Where's the presence work?

PR #1403, opened 2017-06-19
Fixes crash that results from Apple's UICollectionView bug https://openradar.appspot.com/28167779

bakwarte:
This prevents the view controller from crashing when Apple's UICollectionView bug for inserts manifests itself.

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -1268,7 +1268,12 @@ static NSInteger const ATLPhotoActionSheet = 1000;
             for (ATLDataSourceChange *change in objectChanges) {
                 switch (change.type) {
                     case LYRQueryControllerChangeTypeInsert:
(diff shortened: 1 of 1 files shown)


woutergoossens:
Can this be approved? Have this problem with LayerKit 0.26.10 and Atlas 1.1.2

Przytua:
@woutergoossens @bakwarte can you please check v1.1.4 (https://github.com/layerhq/Atlas-iOS/releases/tag/v1.1.4) as this was probably fixed with #1435 ?

magneticrob:
We are also still getting this bug, on the latest version of Atlas and LayerKit. However, we've noticed that it is isolated to when the app is put into the background. Try it a few times in a row if you can't get it to happen straight away.

Steps:

1 - Join a conversation
2 - Put app in background
3 - Send message to that conversation
4 - Observe crash

Issue #1395, opened 2017-05-31
Time stamp got hidden if shouldDisplayUsernameForOneOtherParticipant = false


edwinruan:
**1. How to reproduce?** 
Atlas version : > 1.0.33

set `shouldDisplayUsernameForOneOtherParticipant` to `false`. The date Label in each section will disappear.  
set `shouldDisplayUsernameForOneOtherParticipant` to `true`. The date label will show up with participant username.

**2. Reason:**

The issue is caused by this commit. #1378

In  `ATLConversationViewController`, line 490- 493.  

` if ([self shouldDisplaySenderLabelForSection:indexPath.section]) {`
     `[header updateWithParticipantName:[self participantNameForMessage:message]];`
 `  } else {`
    ` [header updateWithAttributedStringForDate:[[NSAttributedString alloc] initWithString:@"" attributes:nil]];`
  `  }`

In the else condition,  we should  use`[header updateWithParticipantName:""]` since it updates the participant name, not date label.

![time label](https://cloud.githubusercontent.com/assets/14810193/26643487/19b6c7ce-45f7-11e7-9c62-2124de01dfd1.png)

vinzen:
Easy fix ! I don't understand why It has not been fixed yet.

`[header updateWithParticipantName:@" "];` Since `nil` or empty string will display "Unknown User":


```
- (void)updateWithParticipantName:(NSString *)participantName
{
    if (participantName.length) {
        self.participantLabel.text = participantName;
    } else {
        self.participantLabel.text = @"Unknown User";
    }
}
```

daniel-maness:
Fixed in https://github.com/layerhq/Atlas-iOS/releases/tag/v1.1.1

PR #1404, opened 2017-06-20
Fix timestamp when shouldDisplayUsernameForOneOtherParticipant is false

vinzen:
#1395


edwinruan:
Thanks for fixing that issue. :)

PR #1418, opened 2017-08-25
Fixed conversation input field for iOS 11

### LINKED ISSUE #1414: Can't select input field when running iOS 11
Trying to select the input field for a conversation is a NOP on iOS, trying to force first responder is also a NOP. Tested against iOS 11 b1-5, can repeat in the example project.

Przytua:
Fixes #1414 

iOS 11 added ```_UIToolbarContentView``` in UIToolbar, which overlapped the text input, so it didn't respond to tap.


jwyld:
This doesn't do the right thing for the left and right accessory buttons.

Przytua:
@jwyld updated with accessory buttons fixed as well.

PR #1433, opened 2017-11-09
Fixed ATLMessageInputToolbar layout according to safe area insets

### LINKED ISSUE #1426: Incompatibility with the safe area layout guides on iOS 11
This issue affects the iPhone X as the edges of the conversation view and possibly more, exceed the safe areas. 

- In landscape, the top notch covers part of the conversation on either side.
- In portrait, the bottom notch covers the message input bar.

Przytua:
Fixes #1426 

Added ```safeAreaInsets.bottom``` to the ```ATLMessageInputToolbar``` height and subviews during layout so it's not overlapped on iPhone X.

zrzut ekranu 2017-11-09 o 12 38 50zrzut ekranu 2017-11-09 o 12 38 43
zrzut ekranu 2017-11-09 o 12 58 33zrzut ekranu 2017-11-09 o 12 58 37
zrzut ekranu 2017-11-09 o 23 31 13
zrzut ekranu 2017-11-09 o 23 31 16

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -32,6 +32,8 @@
 #import "ATLMediaAttachment.h"
 #import "ATLLocationManager.h"
 #import "LYRIdentity+ATLParticipant.h"
+#import "UIView+ATLHelpers.h"
+#import "UICollectionView+ATLHelpers.h"
 
 @interface ATLConversationViewController () <UICollectionViewDataSource, UICollectionViewDelegate, CLLocationManagerDelegate>
 
@@ -125,10 +127,12 @@ static NSInteger const ATLPhotoActionSheet = 1000;
 {
     [super loadView];
     // Collection View Setup
+    UICollectionViewFlowLayout *flowLayout = [[UICollectionViewFlowLayout alloc] init];
     self.collectionView = [[ATLConversationCollectionView alloc] initWithFrame:CGRectZero
-                                                          collectionViewLayout:[[UICollectionViewFlowLayout alloc] init]];
+                                                          collectionViewLayout:flowLayout];
     self.collectionView.delegate = self;
     self.collectionView.dataSource = self;
+    [self.collectionView atl_setupContentInsetAdjustmentBehavior];
 }
 
 - (void)setLayerClient:(LYRClient *)layerClient
@@ -901,6 +905,15 @@ static NSInteger const ATLPhotoActionSheet = 1000;
 {
     [super willAnimateRotationToInterfaceOrientation:toInterfaceOrientation duration:duration];
     [self.collectionView.collectionViewLayout invalidateLayout];
+    [self.inputAccessoryView setNeedsLayout];
+}
+
+- (void)viewWillTransitionToSize:(CGSize)size
+       withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
+{
+    [super viewWillTransitionToSize:size withTransitionCoordinator:coordinator];
+    [self.collectionView.collectionViewLayout invalidateLayout];
+    [self.inputAccessoryView setNeedsLayout];
 }
 
 #pragma mark - ATLAddressBarViewControllerDelegate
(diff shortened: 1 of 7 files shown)


Code/Views/ATLMessageInputToolbar.m
```diff
@@ -121,6 +121,11 @@ - (void)layoutSubviews
         self.firstAppearance = NO;
     }
     
+    CGFloat bottomInset = 0;
+    if (@available(iOS 11.0, *)) {
```
jwyld: This should really be checking to see `[self respondsToSelector:@selector(safeAreaInsets)]` instead of checking for iOS 11.

Przytua: This throws a warning in Xcode:
zrzut ekranu 2017-11-09 o 18 16 57
This is a new way of checking availability, introduced in Xcode 9 and it removes warnings:
zrzut ekranu 2017-11-09 o 18 17 14

jwyld: That's because it should use `[self performSelector:@selector(safeAreaInsets)]`.

Przytua: This is recommended by Apple ():
> To guard the use of new APIs, use an if statement combined with @available, a new Objective-C expression that checks the system version. For example:
```
if (@available(macOS 10.13, iOS 11, *)) {
    --
    // The compiler will not warn about uses of APIs from macOS 10.13
    // or iOS 11 here
}
```

I believe this is a better solution, as the propery/method call is strongly related to the object and it's type (so it's impossible to use a similarly named selector from another class by mistake), and as well it throws the warning when it's not guarded with ```@available``` so if one would forget to wrap it with if statement, the compiler will remind about it.

Przytua: And of course it's better for refactoring, as if this would be our own class, and we'd rename the method/property, the selector in ```@selector``` might not get changed during renaming.

Code/Controllers/ATLConversationViewController.m
```diff
@@ -1115,9 +1128,10 @@ - (void)notifyDelegateOfMessageSelection:(LYRMessage *)message
     }
 }
 
-- (CGSize)heightForMessageAtIndexPath:(NSIndexPath *)indexPath
+- (CGSize)sizeForMessageAtIndexPath:(NSIndexPath *)indexPath
```
jwyld: I worry about this name change.  Although it's not public, it's still possible developers latched onto the name already.  It's a core method for computing the collection view, so if they did overrides, they likely need to call this method.

Code/Utilities/UICollectionView+ATLHelpers.m
```diff
@@ -0,0 +1,53 @@
...
+//  limitations under the License.
+//
+
+#import "UICollectionView+ATLHelpers.h"
+
+@implementation UICollectionView (ATLHelpers)
+
+- (UIEdgeInsets)atl_adjustedContentInset {
+    SEL selector = NSSelectorFromString(@"adjustedContentInset");
+    if (![self respondsToSelector:selector]) {
```
jwyld: How often does this fire?  If frequent, consider a `dispatch_once` to hold the result.

Code/Utilities/UICollectionView+ATLHelpers.m
```diff
@@ -0,0 +1,53 @@
...
+    [invocation setTarget:self];
+    [invocation invoke];
+    UIEdgeInsets returnValue;
+    [invocation getReturnValue:&returnValue];
+    return returnValue;
+}
+
+- (void)atl_setupContentInsetAdjustmentBehavior {
+    SEL selector = NSSelectorFromString(@"setContentInsetAdjustmentBehavior:");
+    if ([self respondsToSelector:selector]) {
```
jwyld: How often does this fire?  If frequent, consider a `dispatch_once` to hold the result.

Code/Utilities/UIView+ATLHelpers.m
```diff
@@ -0,0 +1,40 @@
...
+//  limitations under the License.
+//
+
+#import "UIView+ATLHelpers.h"
+
+@implementation UIView (ATLHelpers)
+
+- (UIEdgeInsets)atl_safeAreaInsets {
+    SEL selector = NSSelectorFromString(@"safeAreaInsets");
+    if (![self respondsToSelector:selector]) {
```
jwyld: How often does this fire?  If frequent, consider a `dispatch_once` to hold the result.

jwyld:
I'm concerned about all the ruby and gemfile changes.  Is it possible to do this without those?

Przytua:
@jwyld updated according to comments.

jwyld:
Looks good

PR #1435, opened 2017-11-10
Fixed crash in converstaion view controller

### LINKED ISSUE #1164: Crash: ATLConversationViewController, self.collectionView performBatchUpdates!!!
Whenever two participants are exchanging messages rapidly (like 6-7 messages per 5 seconds), this crash happens, The detail of the crash log can be found here. #1159

After spending a lot of time on this I realised that somehow the section count are changing concurrently e.g. at the start of the batch updates the section count is say 60, some how during the batch updates, while exchanging messages rapidly, this count changes (i.e. 62) which triggers NSInternalInconsistencyException.

May be the data source is being updated from multiple sources, I think all the changes to the datasource of collectionView should be done using one single queue of changes.

As a fix to avoid the crash I have enclosed the performBatchUpdate call inside a "@try, @catch", However if anybody can suggest me on how to fix the issue, that would be great help......

### LINKED ISSUE #1230: Changing the list of participants and sending a message at the same time crash the collectionView
I'm sending a message when someone is added or removed from a conversation, and this crash the collection view.

The 2 following methods are trigger:
- configureControllerForChangedParticipants
- queryControllerDidChangeContent

configureControllerForChangedParticipants by calling `[self.collectionView reloadData];`
queryControllerDidChangeContent by calling `[self.collectionView performBatchUpdates:^{`

crash because `- (NSInteger)numberOfSectionsInCollectionView:(UICollectionView *)collectionView`
Does not correspond to the one expected.

Przytua:
Fixes #1230 and fixes #1164

```ATLConversationDataSource``` keeps a copy of messages for presentation in ```ATLConversationViewController```, so whenever ```reloadData``` is called, it uses a shapshot of messages set, instead of all messages currently held by query controller. The copy of messages in ```ATLConversationDataSource``` is updated only during ```performBatchUpdates``` inside ```queryControllerDidChangeContent:``` delegate method, and that prevents the crash.

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -258,6 +258,7 @@ static NSInteger const ATLPhotoActionSheet = 1000;
     }
     self.conversationDataSource.queryController.delegate = self;
     self.queryController = self.conversationDataSource.queryController;
+    [self.conversationDataSource updateMessages];
     [self.collectionView reloadData];
 }
 
@@ -336,7 +337,7 @@ static NSInteger const ATLPhotoActionSheet = 1000;
  */
 - (NSInteger)numberOfSectionsInCollectionView:(UICollectionView *)collectionView
 {
-    return [self.conversationDataSource.queryController numberOfObjectsInSection:0] + ATLNumberOfSectionsBeforeFirstMessageSection;
+    return [self.conversationDataSource numberOfMessages] + ATLNumberOfSectionsBeforeFirstMessageSection;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


jwyld:
Looks good

laposheureux:
Came here from #1403...

@Przytua: We are encountering insert related crashes even with the change in this PR. An insert comes along and upon hitting the batch update, the number of sections remains constant despite a number of insertions, e.g.:

```Invalid update: invalid number of sections. The number of sections contained in the collection view after the update (33) must be equal to the number of sections contained in the collection view before the update (33), plus or minus the number of sections inserted or deleted (2 inserted, 0 deleted).```

magneticrob:
@laposheureux Also still getting this, specifically when the app is in the background.

PR #1440, opened 2017-11-29
Include missing file in order to support carthage

maybeiamme:
In v1.1.4, Atlas is unable to be built with carthage because Atlas has missing files.
So, I added four files to project, and target Atlas
UIView+ATLHelpers.h
UIView+ATLHelpers.m
UICollectionView+ATLHelpers.h
UICollectionView+ATLHelpers.m

Please accept my request.


daniel-maness:
Fixed and merged in #1441

PR #1442, opened 2017-12-07
Fix build issues

cliss:
There are a couple files that are missing from the Xcode project, which is causing a build failure.

* [`UICollectionView+ATLHelpers.h`](https://github.com/layerhq/Atlas-iOS/blob/master/Code/Utilities/UICollectionView%2BATLHelpers.h), added in [this commit](https://github.com/layerhq/Atlas-iOS/commit/a75f57ac63aeb063641d2ad5a0d1547b856b9023)
* [`UIView+ATLHelpers.h`](https://github.com/layerhq/Atlas-iOS/blob/master/Code/Utilities/UIView%2BATLHelpers.h), added in [this commit](https://github.com/layerhq/Atlas-iOS/commit/a75f57ac63aeb063641d2ad5a0d1547b856b9023)

I should also note [`UIMutableNotificationAction+ATLHelpers.h`](https://github.com/layerhq/Atlas-iOS/blob/master/Code/Utilities/UIMutableUserNotificationAction%2BATLHelpers.h) is in the file tree but not referenced anywhere in the project.


cliss:
Taking a second stab at this.

PR #1454, opened 2018-02-28
Location message sent when delegate request that no message is sent

V-FEXrt:
According the the documentation comments above `ATLConversationViewControllerDelegate messagesForMediaAttachments`
```
If `nil` is returned, the controller will fall back to default behavior. If an empty
 `NSOrderedSet` is returned, the controller will not send any messages.
```

However returning an empty set will cause the location message to be set. This PR fixes that so that functionality matches the documentation comment

### DIFF
--- a/Code/Controllers/ATLConversationViewController.m
+++ b/Code/Controllers/ATLConversationViewController.m
@@ -618,8 +618,14 @@ static NSInteger const ATLPhotoActionSheet = 1000;
         return;
     }
     
+    if (self.addressBarController) [self.addressBarController disable];
+    
     // If there's no content in the input field, send the location.
     NSOrderedSet *messages = [self messagesForMediaAttachments:messageInputToolbar.mediaAttachments];
+    
+    // Don't send message if nil
+    if (!messages) return;
+    
     if (messages.count == 0 && messageInputToolbar.textInputView.text.length == 0) {
         [self sendLocationMessage];
     } else {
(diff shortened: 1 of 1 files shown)


cpiggott:
🔥

V-FEXrt:
@daniel-maness  @blakewatters @chipxsd Pinging for visibility

Code/Controllers/ATLConversationViewController.m
```diff
@@ -618,16 +618,21 @@ - (void)messageInputToolbar:(ATLMessageInputToolbar *)messageInputToolbar didTap
         return;
     }
     
+    if (self.addressBarController) [self.addressBarController disable];
+    
     // If there's no content in the input field, send the location.
     NSOrderedSet *messages = [self messagesForMediaAttachments:messageInputToolbar.mediaAttachments];
+    
+    // Don't send message if nil
+    if (!messages) return;
```
T-Pham: `messages` is not nil here but an empty set `{}` so the location is still sent.

T-Pham: Another issue I ran into is that the "Send" button is always enabled after we set a conversation to ATLConversationViewController. And it's because the Send button is enabled, user can just tap it even when the message field is blank, which makes the location being sent.
We can just make sure to disable the "Send" button in case the message field is blank and the problem in the PR would not happen at all because user would not be able to send empty message from the first place. I think it is good to do that [here](https://github.com/layerhq/Atlas-iOS/blob/master/Code/Controllers/ATLConversationViewController.m#L305)

```
BOOL shouldEnableButton = self.conversation && self.messageInputToolbar.textInputView.hasText ? YES : NO;
```

PR #1456, opened 2018-06-13
Prevents messages from being hidden behind navigation bar on iOS 11 f…

daniel-maness:
…or new conversations.


V-FEXrt:
Spent all day yesterday trying to figure this out. 👍

